PR #1, opened 2022-04-19
[SFEQS-686] Scaffold io-functions-sign repository

lucacavallaro:


.eslintrc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "env": {
+    "es2021": true,
+    "node": true
+  },
+  "extends": ["@pagopa/eslint-config/recommended"],
+  "rules": {
+    "functional/immutable-data": "off"
```
gunzip: why off?

.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules
```
gunzip, line 1: I'd add at least 

```
.env*

*.js
!jest.config.js
!.eslintrc.js

```

.node-version
```diff
@@ -0,0 +1 @@
+16.13.0
```
gunzip, line 1: does this match the deployed function runtime version?

docs/openapi.yaml
```diff
@@ -0,0 +1,286 @@
...
+        schema:
+          type: string
+          format: uuid
+    get:
+      operationId: GetSigner
+      tags:
+        - Signer
+      responses:
+        200:
+          description: Il profilo del Signer
```
gunzip: should be in english see https://github.com/pagopa/io-functions-services/blob/master/openapi/index.yaml

host.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": "2.0",
+  "extensionBundle": {
```
gunzip: we usually prefer to control packages version though csjprog

https://github.com/pagopa/io-functions-services/blob/master/extensions.csproj

gunzip (changes requested):

.node-version
lucacavallaro: yes

gunzip (approved):

PR #5, opened 2022-05-26
Modified some endpoints from the user point of view

filippo-tenaglia:
In this first commit I changed the endpoints:
- /signer to manage the parameter that the app will send that should be the fiscal code of the user;
- the qtsp-clauses to manage a proposed structure for the clauses that the app will receive during the onboarding.

For the first call I create a component into openapi-definition.yaml to describe how should be the fiscal code, for the second I created, into the openapi-user.yaml, the component qtsp-clause to manage the single clause.
Tell me if you agree with these changes @gunzip , @lucacavallaro , @hevelius  and @grausof

### DIFF
--- a/openapi-definitions.yaml
+++ b/openapi-definitions.yaml
@@ -123,14 +123,3 @@ FullfilledSignatureRequest:
           type: array
           items:
             $ref: "#/components/schemas/Document"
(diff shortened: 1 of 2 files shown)


gunzip:
obsolete

PR #6, opened 2022-06-03
Implements functions: "RequestSignature", "GetSignatureRequest", "CreateProduct", "GetProduct", "Info"

lucacavallaro:
#### List of Changes

This PR introduces the project's basic structure and implements the following Azure Functions:
- RequestSignature
- GetSignatureRequest
- CreateProduct
- GetProduct
- Info

Works locally, with Azure Functions emulator and Cosmos DB.
The online test is in progress.

#### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [x] I have updated the documentation accordingly.


src/app/use-cases/request-signature.ts
```diff
@@ -0,0 +1,37 @@
...
+  (
+    getSignerByFiscalCode: GetSignerByFiscalCode,
+    addSignatureRequest: AddSignatureRequest
+  ) =>
+  (payload: RequestSignaturePayload): TE.TaskEither<Error, SignatureRequest> =>
+    pipe(
+      getSignerByFiscalCode(payload.fiscalCode),
+      TE.map(
+        (signer): SignatureRequest => ({
+          id: crypto.randomUUID(),
```
gunzip: we generally use `ulid()` which has some nice properties over `uuids`.

https://github.com/ulid/spec

src/signature-request/clause.ts
```diff
@@ -0,0 +1,24 @@
+import * as t from "io-ts";
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+
+export const ClauseTitle = NonEmptyString;
```
gunzip, line 4: i'm wondering if we should take those types from the generated ones from the API specs.

src/app/use-cases/request-signature.ts
```diff
@@ -0,0 +1,37 @@
...
+import { DocumentList } from "../../signature-request/document";
+import { Subscription } from "../../signature-request/subscription";
+import { GetSignerByFiscalCode } from "../../signer/signer";
+
+import {
+  SignatureRequest,
+  AddSignatureRequest,
+} from "../../signature-request/signature-request";
+
+export type RequestSignaturePayload = {
```
gunzip, line 22: I guess this sould be taken from the generated type from the swagger API spec.

openapi-issuer.yaml
```diff
@@ -1,87 +1,284 @@
...
+  /signature-requests/{signatureRequestId}:
     get:
-      operationId: GetSignatureRequest
-      tags:
-        - Signature Request
+      operationId: getSignatureRequestById
+      summary: Get a Signature Request By Id
+      parameters:
+        - in: path
+          name: productId
```
grausof: The parameter` {signatureRequestId}` is not present in parameters list. I think you should replace `productId` with `signatureRequestId`

openapi-issuer.yaml
```diff
@@ -1,87 +1,284 @@
...
+      required:
+        - title
+        - status
+        - detail
+    FiscalCode:
+      type: string
+      description: User's fiscal code.
+      format: FiscalCode
+      x-import: "@pagopa/ts-commons/lib/strings"
+      example: SPNDNL80R13C555X
```
grausof, line 197: @gunzip 😅 I know it's used everywhere, but can't we try to put a mock like AAABBB00A00B000H ?

src/infra/azure/cosmos/product.ts
```diff
@@ -0,0 +1,87 @@
...
+import {
+  Product,
+  AddProduct,
+  GetProduct,
+} from "../../../signature-request/product";
+
+import { container } from "./database";
+
+const containerId = "products";
+const partitionKey = "subscriptionId";
```
grausof, line 25: Why these variables here? It is no better on the environment?

openapi-issuer.yaml
grausof: Fix: https://github.com/pagopa/io-functions-sign/pull/6/commits/f303882e4e2c03772a5e560a3d224ee52c3a9083

src/id.ts
```diff
@@ -0,0 +1,8 @@
+import * as crypto from "node:crypto";
+import * as t from "io-ts";
+
+export const Id = t.string;
+
+export type Id = t.TypeOf<typeof Id>;
+
+export const id = () => crypto.randomUUID();
```
grausof, line 8: Use `ulid()` https://github.com/pagopa/io-functions-sign/pull/6#discussion_r889922839

openapi-issuer.yaml
lucacavallaro: LOL! I didn't notice that it is a real fiscal code: I copied it from `io-functions-services`'s spec

src/infra/azure/cosmos/product.ts
lucacavallaro: Nope, the database schema does not depend on environment (and it is replicated on TF)

The database name, however, can be changed by the `CosmosDbDatabaseName` environment variable.

src/id.ts
lucacavallaro: Sorry, I didn't update the previous comment. 

As I told in our team sync, I am currently doing a comparison between different opaque ID implementations (for example: `ulid`, `xid`, `snowflake` ecc.) so for now we can use this `id` function that in the next PRs will be be updated with the chosen implementation 🙂

grausof: Perfect! ☺️

src/infra/azure/cosmos/config.ts
```diff
@@ -0,0 +1,27 @@
...
+});
+
+export type CosmosConfig = t.TypeOf<typeof CosmosConfig>;
+
+export const getCosmosConfigFromEnvironment: RE.ReaderEither<
+  NodeJS.ProcessEnv,
+  Error,
+  CosmosConfig
+> = sequenceS(RE.Apply)({
+  connectionString: readFromNodeEnv("CosmosDbConnectionString"),
```
hevelius, line 22: Why not define a const instead? `CosmosDbConnectionString` could change?

src/infra/azure/cosmos/config.ts
```diff
@@ -0,0 +1,27 @@
...
+  dbName: pipe(
+    readFromNodeEnv("CosmosDbDatabaseName"),
```
hevelius, line 24: ditto

src/infra/azure/cosmos/product.ts
```diff
@@ -0,0 +1,87 @@
...
+        productModelTE,
+        TE.chain((model) =>
+          pipe(
+            model.create(newProduct),
+            TE.mapLeft(() => new Error("Error creating the product"))
+          )
+        )
+      )
+    )
+  );
```
hevelius, line 69: Would it be possible to avoid nested pipes?

src/infra/azure/cosmos/product.ts
```diff
@@ -0,0 +1,87 @@
...
+        productModelTE,
+        TE.chain((model) =>
+          pipe(
+            model.find([id, subscriptionId]),
+            TE.mapLeft(() => new Error("Error getting the Product"))
+          )
+        )
+      )
+    )
+  );
```
hevelius, line 87: ditto

src/signature-request/__test__/clause.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { isRight as isValid } from "fp-ts/Either";
+import { pipe } from "fp-ts/function";
+import { ClauseTitle, ClauseList } from "../clause";
+
+describe("ClauseTitle", () => {
+  it.each([
+    { title: "", expected: false },
+    { title: "Clause 1", expected: true },
+  ])("should be valid (non empty) ($#)", ({ title, expected }) => {
```
hevelius, line 9: Why not mock this data out of test? ex. in external file `src/signature-request/__mocks__/clause.ts`

src/signature-request/__test__/clause.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    {
+      list: [
+        {
+          title: "Clause 1",
+          required: false,
+        },
+      ],
+      expected: true,
+    },
+  ])("should always have at least one element ($#)", ({ list, expected }) => {
```
hevelius, line 27: ditto

src/ui/functions/get-product.ts
```diff
@@ -0,0 +1,73 @@
...
+import { Product, ProductId } from "../../signature-request/product";
+import { Subscription } from "../../signature-request/subscription";
+
+import { getProduct } from "../../infra/azure/cosmos/product";
+
+export const requireProductId: (
+  req: HttpRequest
+) => E.Either<Error, Product["id"]> = flow(
+  path("productId"),
+  E.fromOption(() => new BadRequestError("Missing productId in path")),
```
hevelius, line 29: Could it be useful to define all error messages in a separate file?

src/infra/azure/cosmos/config.ts
lucacavallaro: These are the names of the environment variables; they aren't utilized anywhere else, so declaring them as constants would be redundant.

src/infra/azure/cosmos/product.ts
lucacavallaro: would you give me an example of how I could change this code?

src/signature-request/__test__/clause.spec.ts
lucacavallaro: as long as the mocked data is small I prefer to keep it in the same file, to make the tests clearer

large mocks will surely be moved in a separate file

src/ui/functions/get-product.ts
lucacavallaro: No, because the error message is tightly related to the code in which it is put and requires no localization (it will always be in English) 

If the same type of error occurs multiple times with the same type of message, I prefer to extract it as a new error (as done with `ProductNotFoundError`)

src/infra/azure/cosmos/product.ts
lucacavallaro: (maybe i can extract some function from this code)

openapi-issuer.yaml
gunzip: You can put anything that satisfy FiscalCode type constraints It's fair to leave it as is for me.

src/ui/api-models/SignatureRequestDetailView.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Do not edit this file it is auto-generated by io-utils / gen-api-models.
```
gunzip, line 2: Should we commit these?

gunzip (approved):

grausof (approved):
just a remark, is it right to use the `lib` folder as `outputDir` in tsconfig? I would use `dist` or `app` instead

lucacavallaro:
> just a remark, is it right to use the `lib` folder as `outputDir` in tsconfig? I would use `dist` or `app` instead

@grausof in NodeJS the de-facto standard for folder naming indicates that the compiled JS code, which the runtime can run "as-is" must be in the `lib` folder

as shown here https://gist.github.com/tracker1/59f2c13044315f88bee9

`dist` is a convention for bundled front-end code, not for back-end in Node

grausof:
@lucacavallaro  Following azure standard, when a function is created in typescript, this is the tsconfig file that is created 
```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es6",
    "outDir": "dist",
    "rootDir": ".",
    "sourceMap": true,
    "strict": false
  }
}
```
https://github.com/microsoft/vscode-azurefunctions/blob/9f72c0cf7849f869d5d52a674463a0a85989efaa/src/constants.ts#L86

src/ui/api-models/SignatureRequestDetailView.ts
lucacavallaro: I'm not sure, so I'd rather leave them committed for now.

Having them in the repository eliminates the need for additional build processes, simplifies project installation, and improves the repository's "explorability."

But it's something I'd like to consider further.

lucacavallaro:
> @lucacavallaro Following azure standard, when a function is created in typescript, this is the tsconfig file that is created
> 
> ```
> {
>   "compilerOptions": {
>     "module": "commonjs",
>     "target": "es6",
>     "outDir": "dist",
>     "rootDir": ".",
>     "sourceMap": true,
>     "strict": false
>   }
> }
> ```

Of course, there are two things to keep in mind:

1) Azure Functions are merely a "detail" of our project's development. They are currently utilized as a "transport layer," but alternative technologies may be utilised in the future. This is a NodeJS project, not a "Azure" project, hence the code is agnostic to Azure. If you notice Azure Functions is mentioned only in `src/ui/functions`, not outside.

2) NodeJS isn't well supported in Azure. Libraries and tutorials are clearly not written by JavaScript developers who are familiar with development standards. The resulting `tsconfig` does not include a `src` folder (which is a huge mistake!). And as a target `es6` is used (which is for a very old version of Node, in Node 16 we use `es2021`) TS libs are also not restricted, according to Azure config we should pull also the DOM API. As a result, we should not use it as a reference.

PR #10, opened 2022-06-30
[SFEQS-846] Add Openapi for QTSP

grausof:
This PR draft adds the openapi specification towards the QTSP following this sequence! 

[Open in swagger!](https://editor.swagger.io/?url=https://raw.githubusercontent.com/pagopa/io-functions-sign/openapi-qtsp-SFEQS-846/docs/openapi-qtsp.yaml)


docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+  schemas:
+  
+    TOS:
+      type: object
+      required:
+        - tos
+      properties:
+        tos:
+          type: array
+          items:
```
gunzip: why array of strings? I'm expecting here at least 1 identifier (ie. an hash value) and a link to the TOS pdf.

It could be something like: `[ { id: <id>,  link: <url>, title: <title> },  ... ]`

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+      type: object
+      required:
+        - tos
+      properties:
+        tos:
+          type: array
+          items: 
+            type: string
+          description: "List of terms of service"
+        privacyLink:
```
gunzip: This could go into the previous array

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/TOS"
+        
+        default:
+          description: Generic error
+      
+      
+      
+  /login:
```
gunzip:
```suggestion
  /challenge:
```

since the operation is called 'GetChallenge' I'd reflect this into the path. Moreover prefer objects to actions (_login_)

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+    post:
+      operationId: GetChallenge
+      tags:
+        - QTSP
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Auth'
```
gunzip:
```suggestion
              $ref: '#/components/schemas/GetChallengeInput'
```

I'd put some reference to "challenge" here since we're not authenticating.

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+    Auth:
+      type: object
+      required:
+        - SAMLAssertion
+        - publicKey
+      properties:
+        SAMLAssertion:
+          type: string
+          format: byte
+          description: "SAML assertion in base64"
```
gunzip: why do we want to send all these informations here? 

I'd send instead: 1) the user fiscal code 2) hashes of the files to be signed

what do you think about it?

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+    SignatureRequest:
+      type: object
+      required:
+        - signedChallenge
+        - tosAccepted
+        - hashToSign
+        - challengeId
+        - personalData
+      properties:
+        tosAccepted:
```
gunzip: this willl probably be some base64 payload with the compiled pdf

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+          type: boolean
+        challengeId:
+          type: string
+          format: uuid
+        signedChallenge:
+          type: string
+          format: byte
+        hashToSign:
```
gunzip:
```suggestion
        hashesToSign:
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+        - email
+      properties:
+        firstName:
+          type: string
+        lastName:
+          type: string
+        fiscalCode:
+          type: string
+        email:
+          type: string
```
gunzip: we don't need to send thse since they're inside the SAML response (apart from the email that we must send aside)

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+  title: Firma con IO QTSP
+  description: Any APIs called by IO Backend to QTSP
+  version: 1.0.0
+
+paths:
+
+  /tos:
+    summary: Request terms of service
+    get:
+      operationId: GetTos
```
lucacavallaro: use `camelCase` in `operationId`

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+          type: string
+        email:
+          type: string
+          
+    SignatureResponse:
+      type: object
+      required:
+        - signedHash
+      properties:
+        signedHash:
```
gunzip: what about the signed timestamp for LTV?

gunzip (changes requested):
can we rewrite these specs using snake_case notation?

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+      responses:
+        '200':
+          description: Terms of service to be accepted
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/TOS"
+        
+        default:
+          description: Generic error
```
lucacavallaro: you should add details about the"shape" of the error response (for example we can use the format described in [RFC7807](https://datatracker.ietf.org/doc/html/rfc7807), which we use in all our APIs)

docs/openapi-qtsp.yaml
lucacavallaro: What is the purpose of this endpoint?
1. the `summary` is vague
2. does exists a resource called `Login` ?

in the rest API, the endpoints must always contain the name of the `resource` that are the subject of the request, so if `login` here is used as a `verb` it should be changed with a more adequate name

here you can find an explanation:  https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#restful

grausof:
> can we rewrite these specs using snake_case notation?

@gunzip In the [other specifications](https://github.com/pagopa/io-functions-sign/blob/main/openapi-issuer.yaml) the camelCase was used. Do you think we will have to change them?

docs/openapi-qtsp.yaml
grausof: why a link for each TOS? I expect there to be only a final PDF with a single link while the TOS will be strings to be displayed as a bulleted list. The TOS to show could also support the markdown maybe.

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+        400:
+          description: Bad request
+        401:
+          description: SAML assertion not verified
+        404:
+          description: User not found
+        default:
+          description: Generic error
+      
+  /sign:
```
lucacavallaro: `sign` is a verb, so there is the same problem as the `login` endpoint

is this the endpoint that actually makes the `signature` ?

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+          description: Challenge not found
+        default:
+          description: Generic error
+
+        
+          
+components:
+  schemas:
+  
+    TOS:
```
lucacavallaro: I think it should be an array of objects, with this shape I don't understand what the resource is

docs/openapi-qtsp.yaml
grausof: To request the challenge we must necessarily send SAMLAssertion and publicKey. The data relating to the fiscal code and the hashes to be signed will be sent together with the signed challenge. If the QTSP still has to generate the challenge, it makes no sense to send the file hashes.

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,185 @@
...
+        - challengeToSign
+      properties:
+        challengeId:
+          type: string
+          format: uuid
+        challengeToSign:
+          type: string
+          format: byte
+          
+    SignatureRequest:
```
lucacavallaro:
```suggestion
    CreateSignatureBody:
```

docs/openapi-qtsp.yaml
lucacavallaro: here we should put a reference to each accepted term, in my opinion (a boolean is not enough)

lucacavallaro (changes requested):

docs/openapi-qtsp.yaml
gunzip: Ok, now I get it. But I'm missing the link to this PDF here. I'd also expect to have `title` and `id` for this PDF.

docs/openapi-qtsp.yaml
gunzip: Why must we "necessarily send SAMLAssertion and publicKey" ? I'm suggesting to send the hashes since in my view
the challenge is related to the tuple (user -> files to be signed).

docs/openapi-qtsp.yaml
lucacavallaro: Even if the terms of service document were unique (which is unsure), the QTSP could use anchors to refer to the appropriate section for each term.

Additionally, as I noted in my review remark, this response does not make it clear which `resource` is the actual focus of the request.

docs/openapi-qtsp.yaml
grausof: we only need the signed hash. All other information can be gleaned. For the timestamp it is necessary to have the url of a provider with RFC3161 standard

docs/openapi-qtsp.yaml
grausof: I understand, what do you think of a structure like this?
```
"tos": [
    {
      "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
      "title": "string",
      "link": "string",
      "sections": [
        {
          "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
          "description": "string",
          "anchor": "string"
        }
      ]
    }
  ]
```

docs/openapi-qtsp.yaml
grausof: so we will also send a pre-filled PDF? Wouldn't it be better to just send the list of accepted TOS ids?

docs/openapi-qtsp.yaml
grausof: I know. For now I had decided not to include them so as not to burden the reading of the yaml.

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,283 @@
...
+    post:
+      operationId: getChallenge
+      tags:
+        - QTSP
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateChallengeBody'
```
gunzip: get or create?

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,283 @@
...
+    post:
+      operationId: getSignature
+      tags:
+        - QTSP
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateSignatureBody'
```
gunzip: get or create?

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,283 @@
...
+                $ref: "#/components/schemas/Challenge"
+        "400":
+          $ref: "#/components/responses/BadRequest"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        default:
+          $ref: "#/components/responses/Unexpected"
+      
+  /signature:
+    summary: Send a signature request
```
gunzip:
```suggestion
    summary: Get signatures for the provided hashes
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,283 @@
...
+        - QTSP
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateSignatureBody'
+      responses:
+        '200':
+          description: Challenge generated
```
gunzip:
```suggestion
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,283 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SignatureResponse"
+        "400":
+          $ref: "#/components/responses/BadRequest"
+        "401":
+          description: Challenge signature invalid
```
gunzip, line 77:
```suggestion
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,283 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ProblemJson"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          description: Challenge not found
```
gunzip, line 85:
```suggestion
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,283 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ProblemJson"
```
gunzip: Since it's just "not found" we can omit the payload here.

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,283 @@
...
+      description: Digest of a file.
+      format: byte
+      example: uU0nuZNNPgilLlLX2n2r+sSE7+N6U4DukIj3rOLvzek=
+
+    Challenge:
+      type: object
+      required:
+        - challengeToSign
+      properties:
+        challengeToSign:
```
gunzip:
```suggestion
        challenge:
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,283 @@
...
+    CreateSignatureBody:
+      type: object
+      required:
+        - tosAccepted
+        - signedChallenge
+        - SAMLAssertion
+        - publicKey
+        - fiscalCode
+        - hashesToSign
+        - userEmail
```
gunzip: snake_case

gunzip (changes requested):

docs/openapi-qtsp.yaml
grausof: description is mandatory

docs/openapi-qtsp.yaml
grausof: description is mandatory

docs/openapi-qtsp.yaml
grausof: description is mandatory

docs/openapi-qtsp.yaml
grausof: OK

docs/openapi-qtsp.yaml
lucacavallaro: the resource is `TOS` so the response of `/tos` should contain the array of the required TOS

```json
[
  {
    "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
    "title": "string",
    "link": "string",
  },
]
```

docs/openapi-qtsp.yaml
gunzip: that's right but it looks like it's wrong (isn't this a signature?)

grausof: Yes you are right. Here I have changed the description

docs/openapi-qtsp.yaml
lucacavallaro: 1. If the requested resource is `Challenge` the name of the endpoint must be `challenges` (in plural)
2. A `POST` request in REST is always a "create"
3. The status code on success should be `201`

lucacavallaro (changes requested):

docs/openapi-qtsp.yaml
grausof: Mhhh I thought create as the server must generate the challenge and it does not yet exist at the time of the request. But maybe we can follow the standard and use get?

docs/openapi-qtsp.yaml
grausof: Yes, now it is. the comment refers to a previous implementation.

docs/openapi-qtsp.yaml
```diff
@@ -40,7 +40,7 @@ paths:
             schema:
               $ref: '#/components/schemas/CreateChallengeBody'
```
gunzip: there's still some disalignment: the operation id is `getChallenge` but the payload is named `CreateChallenge` (choose one : - )

docs/openapi-qtsp.yaml
```diff
@@ -66,8 +66,8 @@ paths:
             schema:
               $ref: '#/components/schemas/CreateSignatureBody'
```
gunzip: ditto

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,279 @@
...
+              schema:
+                $ref: "#/components/schemas/Challenge"
+        "400":
+          $ref: "#/components/responses/BadRequest"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        default:
+          $ref: "#/components/responses/Unexpected"
+      
+  /signature:
```
lucacavallaro: `/signatures`

docs/openapi-qtsp.yaml
```diff
@@ -232,52 +228,52 @@ components:
...
           $ref: "#/components/schemas/FiscalCode"
-        hashesToSign:
+        hashes_to_sign:
           type: array
           items: 
             $ref: "#/components/schemas/DocumentHash"
-        userEmail:
+        user_email:
           type: string
           format: byte
```
gunzip: Should be `EmailString` from io-ts-commons

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,279 @@
...
+    TermsOfService:
+      type: object
+      required:
+        - tos
+      properties:
+        tos:
+          type: array
+          items: 
+            $ref: "#/components/schemas/TermOfService"
+          description: "List of terms of service"
```
lucacavallaro:
```suggestion
    TermsOfService:
      type: array
      items: 
         $ref: "#/components/schemas/TermOfService"
      description: "List of terms of service"
```

lucacavallaro (changes requested):

gunzip (changes requested):

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,273 @@
...
+                $ref: "#/components/schemas/Challenge"
+        "400":
+          $ref: "#/components/responses/BadRequest"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        default:
+          $ref: "#/components/responses/Unexpected"
+
+  /signatures:
+    summary: Get signatures for the provided hashes
```
gunzip:
```suggestion
    summary: Request signatures for the provided hashes
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,273 @@
...
+    post:
+      operationId: createSignature
+      tags:
+        - QTSP
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/CreateSignatureBody"
```
gunzip:
```suggestion
              $ref: "#/components/schemas/RequestSignaturesBody"
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,273 @@
...
+      responses:
+        "201":
+          description: Signature hash generated
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SignatureResponse"
```
gunzip:
```suggestion
                $ref: "#/components/schemas/RequestSignaturesResponse"
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,273 @@
...
+          $ref: "#/components/responses/BadRequest"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        default:
+          $ref: "#/components/responses/Unexpected"
+
+  /signatures:
+    summary: Get signatures for the provided hashes
+    post:
+      operationId: createSignature
```
gunzip:
```suggestion
      operationId: requestSignatures
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,273 @@
...
+          $ref: "#/components/responses/BadRequest"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        default:
+          $ref: "#/components/responses/Unexpected"
+
+  /challenges:
+    summary: Request a challenge to QTSP
+    post:
+      operationId: createChallenge
```
gunzip:
```suggestion
      operationId: requestChallenge
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,273 @@
...
+      tags:
+        - QTSP
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/CreateChallengeBody"
```
gunzip:
```suggestion
              $ref: "#/components/schemas/RequestChallengeBody"
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,273 @@
...
+          format: string
+        description:
+          type: string
+          format: string
+        link:
+          type: string
+          format: uri
+          description: "Link to TOS"
+
+    TermsOfServiceAccepted:
```
gunzip:
```suggestion
    TermOfServiceAccepted:
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,273 @@
...
+      type: string
+      format: uuid
+
+    CreateChallengeBody:
```
gunzip:
```suggestion
    RequestChallengeBody:
```

docs/openapi-qtsp.yaml
```diff
@@ -0,0 +1,273 @@
...
+          type: array
+          items:
+            $ref: "#/components/schemas/DocumentHash"
+        user_email:
+          type: string
+          format: EmailString
+          x-import: "@pagopa/ts-commons/lib/strings"
+          example: example@domain.com
+
+    SignatureResponse:
```
gunzip:
```suggestion
    RequestSignaturesResponse:
```

gunzip (approved):

docs/openapi-qtsp.yaml
gunzip: I think that the format proposed by @grausof could satisfy all use cases (clauses tied to specific documents)

docs/openapi-qtsp.yaml
grausof: and maybe add an optional description field for the document!

docs/openapi-qtsp.yaml
grausof: and maybe add an optional description field for the document

lucacavallaro (approved):

lucacavallaro:
Bravo @grausof  💪

PR #12, opened 2022-07-12
Create a new ADR to explain why the Undici library is the best to imp…

filippo-tenaglia:
This ADR contains why we will use Undici to manage the calls to external API

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/architecture/decisions/0004-choose-http-client-library.md
+++ b/docs/architecture/decisions/0004-choose-http-client-library.md
@@ -1,31 +1,30 @@
-# 1. ADR title (choice to make)
+# 1. Choose the best HTTP Client library on Node JS
 
-Date: YYYY-MM-DD
+Date: 2022-07-12
 
 ## Status
 
-Accepted / Proposed / Deprecated / Superseded
+Accepted
 
 ## Context
 
-Describe which problem are you trying to solve and the possibile 
-solutions in terms of UX, Scalability, Autonomy, Adaptability,Security,
-Privacy, Simplicity.
+We must choose a library to implements the external API call that we will develop.
 
 ### Option 1
 
-Describe option 1
+Undici ( https://github.com/nodejs/undici ()
 
-### Option 2
+###Option 2
+Fetch ( https://www.npmjs.com/package/node-fetch )
 
-Describe option 2
+### Option 3
+
+AXIOS ( https://axios-http.com/docs/intro )
 
 ## Decision
 
-In the context of <use case/user story u>, facing <concern c> we decided for <option o> and neglected <other options>
+In the context of Firma Con IO, facing against the the implements of call of http external API (backend IO, Tokenizer and QTSP) we decided for Undici because:
 
-## Consequences
-
-To achieve <system qualities/desired consequences>
-we accept <downside d/undesired consequences>
-because <additional rationale>.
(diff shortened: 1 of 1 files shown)


grausof:
@filippo-tenaglia the file contains only an empty template!

grausof (approved):
Ok for me!

docs/architecture/decisions/0004-choose-http-client-library.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Context
+
+We must choose a library to implements the external API call that we will develop.
+
+### Option 1
+
+Undici ( https://github.com/nodejs/undici ()
+
+###Option 2
```
grausof:
```suggestion
### Option 2
```

docs/architecture/decisions/0004-choose-http-client-library.md
```diff
@@ -0,0 +1,30 @@
...
+Fetch ( https://www.npmjs.com/package/node-fetch )
+
+### Option 3
+
+AXIOS ( https://axios-http.com/docs/intro )
+
+## Decision
+
+In the context of Firma Con IO, facing against the the implements of call of http external API (backend IO, Tokenizer and QTSP) we decided for Undici because:
```
grausof:
```suggestion
In the context of Firma Con IO, facing against the implements of call of http external API (backend IO, Tokenizer and QTSP) we decided for Undici because:
```

docs/architecture/decisions/0004-choose-http-client-library.md
```diff
@@ -0,0 +1,30 @@
...
+
+1. it is an official evolution of Fetch;
+2. it follow the javascritp standards;
+   3( it will be include directly on Node.JS in v. 18)
```
grausof:
```suggestion
3. it will be include directly on Node.JS in v. 18
```

grausof (changes requested):

docs/architecture/decisions/0004-choose-http-client-library.md
```diff
@@ -0,0 +1,31 @@
+# 1. Choose the best HTTP Client library on Node JS
```
gunzip, line 1: A couple of warnings:

1. since the http(s) calls that a function can make in parallel are limited, in the whole IO codebase we rely on a [specific implementation of Http(s) agent](https://github.com/pagopa/io-ts-commons/blob/master/src/agent.ts) that uses node-fetch to set up the right **keepalive** settings taking values from the environment; without this setup the function app will starve. if you want to use an alternative you need to ensure that the correct agent is passed on boostrap (backporting that implementation)
1. you need to ensure that your fetch calls have a **timeout**; in order to do that we use a custom implementation of an Abortable fetch that uses the Abort agent, see for example here: https://github.com/pagopa/pagopa-checkout-fe/blob/main/src/utils/config/fetch.ts#L35

cc @lucacavallaro @grausof

grausof: Thank you! In the final implementation I will take these warnings into consideration!

grausof:
I close this PR because based on @gunzip's comment we need to use some features that are still experimental in `undici` and therefore we will continue to use `node-fetch`.

PR #13, opened 2022-07-14
[SFEQS-603] Add expiryAt field to the signature request

grausof:
This PR adds an optional field `expiryAt` to the signature request to indicate the expiration date and time.
The `expiryAt` field is validated with a check on the date if it is future compared to the current date.

Also mark files in src/ui/api-models/* as autogenerated with [linguist-generated=true](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)


openapi-issuer.yaml
```diff
@@ -309,6 +312,8 @@ components:
           $ref: "#/components/schemas/ProductId"
         fiscalCode:
           $ref: "#/components/schemas/FiscalCode"
+        expiryAt:
```
gunzip:
```suggestion
        expiresAt:
```

I suppose that the API should be snake case (it could be fixed with another PR)

src/signature-request/expiration-datetime.ts
```diff
@@ -0,0 +1,26 @@
+import * as t from "io-ts";
+
+import * as E from "fp-ts/lib/Either";
+import { pipe } from "fp-ts/lib/function";
+
+import { isBefore } from "date-fns";
+
+const isDate = (m: unknown): m is Date => m instanceof Date;
+export const FutureDate = new t.Type<Date, string, unknown>(
```
gunzip: I think that the "is" part should handle the expiration as well otherwise 
`FutureDate.is(<some time in the past>)` will pass (can you add some unit tests about this?)

probably this type is overkill (moreover, it's a little bit odd it changes in time) and an inline check with `isBefore` suffices.

gunzip (changes requested):

openapi-issuer.yaml
gunzip: What about `validBefore` instead?

src/signature-request/expiration-datetime.ts
grausof: I agree. Better not have this type, I'll change the validation

src/app/use-cases/request-signature.ts
```diff
@@ -39,10 +42,22 @@ export const makeRequestSignature =
           TE.chainW(TE.fromOption(() => productNotFoundError)),
           TE.chainEitherK(getDocumentsByMetadata)
         ),
+        expiresAt: pipe(
```
gunzip, line 45: are you sure you need a pipe (I see only one argument).

src/app/use-cases/request-signature.ts
```diff
@@ -39,10 +42,22 @@ export const makeRequestSignature =
...
+          payload.expiresAt
+            ? isBefore(payload.expiresAt, new Date())
+              ? TE.left(
+                  new InvalidDateError(
+                    "The expiration date is earlier than the current one"
```
gunzip:
```suggestion
                    "The expiration date must be in the future"
```

src/app/use-cases/request-signature.ts
```diff
@@ -39,10 +42,22 @@ export const makeRequestSignature =
...
+                  )
+                )
+              : TE.right(payload.expiresAt)
+            : TE.right(undefined)
```
gunzip: can you add unit tests to `makeSignatureRequest`?

grausof: Oops! 😬 Typo

src/app/use-cases/request-signature.ts
grausof: Oops! 😬 Typo

src/app/use-cases/__test__/request-signature.spec.ts
```diff
@@ -0,0 +1,90 @@
+import { pipe } from "fp-ts/function";
+import { isRight as isValid } from "fp-ts/Either";
```
gunzip: please don't rename this kind of methods otherwise it makes code harder to be read 
(I'd to look back for what `isValid` stands for :- ) just `import * from E` and use `E.isRight`

src/app/use-cases/__test__/request-signature.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    const makeRequest = pipe(
+      payload as RequestSignaturePayload,
+      makeRequestSignature(
+        mockGetSignerByFiscalCode,
+        mockGetProduct,
+        mockAddSignatureRequest
+      )
+    )();
+    return makeRequest.then((data) =>
+      expect(pipe(data, isValid)).toBe(expected)
```
gunzip: what about checking if it returns the right error for invalid cases?

src/app/use-cases/__test__/request-signature.spec.ts
grausof: 😬 Sorry, I'll fix it immediately. I believe VS code has done this!

grausof: Sorry, I'll fix it immediately. I inherited this syntax from [here](https://github.com/pagopa/io-functions-sign/blob/6bb7181fce66f4faeca07cbc3d441968efa2a690/src/signature-request/__test__/document.spec.ts#L1)

src/app/use-cases/request-signature.ts
```diff
@@ -39,10 +42,20 @@ export const makeRequestSignature =
...
         ),
+        expiresAt: payload.expiresAt
+          ? isBefore(payload.expiresAt, new Date())
+            ? TE.left(
+                new InvalidEntityError(
+                  "The expiration date must be in the future"
+                )
+              )
+            : TE.right(payload.expiresAt)
+          : TE.right(undefined),
```
gunzip:
```suggestion
        expiresAt: pipe(
          TE.of(payload.expiresAt),
          TE.filterOrElse(
            (_) => _ !== undefined && isBefore(_, Date.now()),
            () =>
              new InvalidEntityError(
                "The expiration date must be in the future"
              )
          )
        ),
```

gunzip: you may consider to extract the predicate lambda into it's own method or to revert to its own type (FutureDate) to make the pipe terser ie. 

```
 expiresAt: pipe(
          TE.of(payload.expiresAt),
          TE.filterOrElse(
            FutureDate.is, // or isDateInTheFuture
            constant(
              new InvalidEntityError(
                "The expiration date must be in the future"
              ))
          )
        ),
```

grausof: but `payload.expiresAt` could also be **undefined** (since it is optional). Could it be correct to write such a thing?
```
export function isFutureDate(date: UTCISODateFromString | undefined) {
  return date === undefined || isAfter(date, new Date());
}
```

gunzip: sorry you're right, just keep the check inline:

```
        expiresAt: pipe(
          TE.of(payload.expiresAt),
          TE.filterOrElse(
            (_) => _ === undefined || isAfter(_, Date.now()),
            constant(
              new InvalidEntityError(
                "The expiration date must be in the future"
              )
            )
          )
        ),

```

gunzip (approved):

PR #16, opened 2022-08-02
[SFEQS-791] Send the signature requests on app IO by messages 

grausof:
#### List of Changes
- Sending a signature request on `Storage Queue` when it is ready to be signed (all documents have been uploaded).
- Trigger of the `SendSignatureRequestMessage` function from the `Storage Queue`
- Sending a message to the IO app via a _standard service._
- Update to latest `handler-kit` API

The project works locally and it is possible to try every part of it (thanks to `azurite`).
A test standard service was used to send the messages.

#### Types of changes

#### Checklist:


README.md
```diff
@@ -38,6 +38,21 @@ And run the project with
 npm start
 ```
 
+### Message services integration
+
+To use a message sending service to the IO app it is also necessary to set the following parameters in `local.settings.json`
```
gunzip: Since these look like local development instructions, it's useless to point to the real IO API since the IO App used for local development talks to the fake backend set up locally using docker. I

Ideally this service (io-functions-sign) should be run inside io-mock (the IO backend simulator) when you want to test the integration with other services. By now, I'd remove these instructions here (these var will be setup inside docker-compose anyway).

gunzip: How did you test the integration?

README.md
```diff
@@ -97,3 +113,19 @@ To accomplish this, simply add the following property to the personal `settings.
 ```
 
 Please care of this setting, as it exposes the editor to potential security issues with third-party extensions.
+
+### Storage integration
+
+A storage account must also be configured to use blob storage and queues in `local.settings.json`
```
gunzip: Instead of documenting environment variables inside the readme, it's better to have a .env file with inline comments (just place comments above the variables). That's the .env included by docker-compose.

SendSignatureRequestMessage/function.json
```diff
@@ -0,0 +1,12 @@
...
+  "bindings": [
+    {
+      "name": "sendSignatureRequestMessage",
+      "type": "queueTrigger",
+      "direction": "in",
+      "queueName": "waiting-for-signature",
+      "connection": "StorageAccountConnectionString"
+    }
+  ],
+  "scriptFile": "../lib/infra/azure/functions/send-signature-request-message.js"
```
gunzip, line 11: The path here should be `../lib/ui/functions/send-signature-request-message.js`

package.json
```diff
@@ -11,7 +11,8 @@
     "prepare": "husky install",
     "func:ext-install": "func extensions install",
     "start": "func start",
-    "generate:issuer-api-models": "gen-api-models --api-spec ./openapi-issuer.yaml --out-dir ./src/ui/api-models"
+    "generate:issuer-api-models": "gen-api-models --api-spec ./openapi-issuer.yaml --out-dir ./src/ui/api-models",
+    "generate:message-api-models": "gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/io-functions-services/master/openapi/index.yaml --out-dir ./src/ui/api-models"
```
gunzip: This is a little controversial: if you point to the master branch (instead of a particular release) you get non reproducible builds. I'd use a specific releae in the path here. 

What about it @lucacavallaro ?

Ideally every service should publish it's SDK (API client) and other projects should use that one (handled with package.json) instead of generating the client locally. There's already a pipeline ready to do this, but I don't know why it's not used yet cc @gquadrati :-)

src/app/use-cases/send-signature-request.ts
```diff
@@ -0,0 +1,98 @@
...
+    );
+
+export const sendMessage =
+  (message: NewMessage, signerId: string) =>
+  (
+    getFiscalCodeBySignerId: GetFiscalCodeBySignerId
+  ): TE.TaskEither<Error, CreatedMessage> =>
+    pipe(
+      sequenceS(TE.ApplySeq)({
+        signer_cf: getFiscalCodeBySignerId(signerId),
```
gunzip:
```suggestion
        signerFiscalCode: getFiscalCodeBySignerId(signerId),
```

src/app/use-cases/send-signature-request.ts
```diff
@@ -0,0 +1,98 @@
...
+      TE.map(({ product }) => mockMakeMessage(signatureRequest, product))
+    );
+
+export const sendMessage =
+  (message: NewMessage, signerId: string) =>
+  (
+    getFiscalCodeBySignerId: GetFiscalCodeBySignerId
+  ): TE.TaskEither<Error, CreatedMessage> =>
+    pipe(
+      sequenceS(TE.ApplySeq)({
```
gunzip: Why are you using sequence here with just one parameter? you can just use chain.

src/app/use-cases/send-signature-request.ts
```diff
@@ -0,0 +1,98 @@
...
+  (sendSignatureRequestBody: SendSignatureRequestBody) =>
+    pipe(
+      getSignatureRequest(
+        sendSignatureRequestBody.signatureRequestId,
+        sendSignatureRequestBody.subscriptionId
+      ),
+      TE.chainW(
+        flow(
+          TE.fromOption(() => signatureRequestNotFoundError),
+          TE.chain((sr) =>
```
gunzip:
```suggestion
          TE.chain((signatureRequest) =>
```

please use meaningful names

src/app/use-cases/send-signature-request.ts
gunzip: Anyway, I won't make sendMessage dependant on `GetFiscalCodeBySignerId`, just call this method outside it and provide the `fiscalCode` directly (I guess you don't need this method at all since you can just pipe `submitMessageForUser(signerFiscalCode)`)

src/app/use-cases/send-signature-request.ts
```diff
@@ -0,0 +1,98 @@
...
+      signatureRequest.expiresAt ? signatureRequest.expiresAt : "never"
+    }\`\n- docs: \`${JSON.stringify(signatureRequest.documents)}\`\n `,
+  },
+});
+
+export const prepareMessage =
+  (getProduct: GetProduct) =>
+  (signatureRequest: SignatureRequest): TE.TaskEither<Error, NewMessage> =>
+    pipe(
+      sequenceS(TE.ApplySeq)({
```
gunzip: Just use chain when you have only one parameter (instead of sequence)

src/app/use-cases/validate-document.ts
```diff
@@ -68,5 +71,10 @@ export const makeValidateDocument =
         )
       ),
       TE.chain(upsertSignatureRequest),
+      TE.chain((sr) =>
```
gunzip:
```suggestion
      TE.chain((signatureRequest) =>
```

src/infra/azure/functions/send-signature-request-message.ts
```diff
@@ -0,0 +1,50 @@
+import * as t from "io-ts";
```
gunzip, line 1: this file must be inside `src/lib/functions`

src/infra/azure/functions/send-signature-request-message.ts
```diff
@@ -0,0 +1,50 @@
...
+);
+
+export const SendSignatureRequestBody = t.type({
+  signatureRequestId: SignatureRequestId,
+  subscriptionId: SubscriptionId,
+});
+export type SendSignatureRequestBody = t.TypeOf<
+  typeof SendSignatureRequestBody
+>;
+
```
gunzip, line 50: may you put a comment on what this method is supposed to do?

src/infra/azure/functions/validate-document-upload.ts
```diff
@@ -40,10 +45,27 @@ const isDocumentUploadedToBlobStorage = pipe(
   )
 );
 
+const enqueueRequestAwaitingSignature = (request: SignatureRequest) =>
```
gunzip, line 52: please add a comment stating whatt this method is supposed to do

src/infra/azure/functions/validate-document-upload.ts
```diff
@@ -9,7 +9,10 @@ import { last } from "fp-ts/ReadonlyNonEmptyArray";
 import { split } from "fp-ts/string";
```
gunzip, line 9: all function handlers must go into `src/lib/functions`

src/infra/azure/storage/queue/client.ts
```diff
@@ -0,0 +1,24 @@
+import { QueueServiceClient, QueueClient } from "@azure/storage-queue";
+
+import * as TE from "fp-ts/TaskEither";
+
+export const createQueueClient = (
```
gunzip: clients must be singletons shared by all functions, so what you realy want here is to export the instance directly (not a factory method)

see the `database` method into `database.ts`

src/infra/azure/storage/queue/client.ts
```diff
@@ -0,0 +1,24 @@
...
+) => {
+  const queueServiceClient =
+    QueueServiceClient.fromConnectionString(connectionString);
+  return queueServiceClient.getQueueClient(queueName);
+};
+
+export const enqueueMessage = (queueClient: QueueClient) => (message: string) =>
+  TE.tryCatch(
+    () => queueClient.sendMessage(Buffer.from(message).toString("base64")),
+    () => new Error("Message cannot be queued")
```
gunzip:
```suggestion
    toError
```

if you swallows errors this way we cannot have any info on why the error occurred

src/infra/azure/storage/queue/signature-request.ts
```diff
@@ -0,0 +1,23 @@
...
+import { pipe } from "fp-ts/function";
+
+import { SignatureRequest } from "../../../../signature-request/signature-request";
+import { enqueueMessage } from "./client";
+
+export type EnqueueSignatureRequest = (
+  request: SignatureRequest
+) => TE.TaskEither<Error, SignatureRequest>;
+
+export const enqueueSignatureRequest =
```
gunzip: since there's nothing specific to azure here, I guess this file should be inside the use-case directory (the business logic goes there), isn't it @lucacavallaro ?

src/infra/services/config.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as E from "fp-ts/Either";
+import * as RE from "fp-ts/ReaderEither";
+import { pipe, flow } from "fp-ts/function";
+
+import { sequenceS } from "fp-ts/lib/Apply";
+
+import { IntegerFromString } from "@pagopa/ts-commons/lib/numbers";
+import { readFromNodeEnv } from "../../app/env";
+
+export const ServiceConfig = t.type({
```
gunzip: I'd rename everything to something like 'ServiceClient' or (better) 'IOApiClient'  since `Service` is too generic.

src/infra/services/send-message.ts
```diff
@@ -0,0 +1,59 @@
...
+      TE.chain(({ basePath, headers, timeoutFetch, path }) =>
+        pipe(
+          TE.tryCatch(
+            () =>
+              timeoutFetch(`${basePath.href}${path.href}`, {
+                body: JSON.stringify(body),
+                method: "POST",
+                headers,
+              })
+                .then((r) => {
```
gunzip: it's better to don't mix Promise code with TaskEithers, just use map / filter ops.

src/infra/services/send-message.ts
```diff
@@ -0,0 +1,59 @@
...
+                  if (!is2xx(r)) {
+                    throw Error("Unexpected webhook response");
+                  }
+                  return r;
+                })
+                .then((r) => r.json()),
```
gunzip: ditto

src/infra/services/send-message.ts
```diff
@@ -0,0 +1,59 @@
...
+            E.toError
+          )
+        )
+      ),
+      TE.chain(
+        flow(
+          CreatedMessage.decode,
+          E.mapLeft(() => new Error("Invalid response")),
```
gunzip: please don't swallow errors, use `readableReport` method to output meaningful decoding errors instead. probably this should be handled at top level (inside handler-kit?): if the bubbled error is ValidationErrors output them with a human readable format

what about it @lucacavallaro ?

src/infra/services/service.ts
```diff
@@ -0,0 +1,47 @@
...
+} from "@pagopa/ts-commons/lib/fetch";
+
+import * as E from "fp-ts/lib/Either";
+
+import { UrlFromString } from "@pagopa/ts-commons/lib/url";
+import { pipe } from "fp-ts/lib/function";
+import { Millisecond } from "@pagopa/ts-commons/lib/units";
+import { config } from "../../app/config";
+
+export const basePath = pipe(
```
gunzip: what about using the generated API client instead?

src/infra/services/service.ts
```diff
@@ -0,0 +1,47 @@
...
+  config,
+  E.chain((config) =>
+    pipe(
+      config.service.serviceBasePath,
+      UrlFromString.decode,
+      E.mapLeft(() => new Error("Invalid service parameters"))
```
gunzip: don't swallow decoding errors

src/ui/api-models/Activation.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
gunzip: I'm not sure these generated files should be committed. Otherwise, every time the IO Api changes, we'll commit unrelated changes when pushgin PRs. What about it @lucacavallaro ?

gunzip (dismissed):

package.json
gquadrati: It's true, we have it https://www.npmjs.com/package/@pagopa/io-functions-services-sdk
If you don't need features in preview, you can use it

gunzip: Just check fp-ts and io-ts required versions since they're a dependency of the SDK as well @grausof

grausof: @gquadrati where can I find the documentation?

gquadrati: @grausof we are a bit lacking on specs :D 
But this is basically a package containing the same client and models you're generating right now.

src/infra/azure/storage/queue/signature-request.ts
grausof: It is here because it depends on the `QueueClient` which is relative to `azure`

src/infra/services/send-message.ts
grausof: Sorry, here I forgot to use `validate`

grausof: Fixed with: 
```
TE.chainW(
        flow(validate(CreatedMessage, "Invalid response"), TE.fromEither)
      )
```

src/app/use-cases/send-signature-request.ts
```diff
@@ -0,0 +1,79 @@
...
+                  signatureRequest.signerId,
+                  getFiscalCodeBySignerId,
+                  TE.chain(submitMessageForUser(message))
+                )
+              )
+            )
+          )
+        )
+      ),
+      TE.map(() => void 0)
```
gunzip:
```suggestion
```

src/app/use-cases/send-signature-request.ts
```diff
@@ -0,0 +1,79 @@
...
+export const prepareMessage =
+  (getProduct: GetProduct) =>
+  (signatureRequest: SignatureRequest): TE.TaskEither<Error, NewMessage> =>
+    pipe(
+      getProduct(signatureRequest.productId, signatureRequest.subscriptionId),
+      TE.chainW(TE.fromOption(() => productNotFoundError)),
+      TE.map((product) => mockMakeMessage(signatureRequest, product))
+    );
+
+export const sendSignatureRequest =
```
gunzip, line 48: may you add a comment on what this method does?

src/app/use-cases/validate-document.ts
```diff
@@ -68,5 +71,10 @@ export const makeValidateDocument =
         )
       ),
       TE.chain(upsertSignatureRequest),
+      TE.chain((signatureRequest) =>
+        status(signatureRequest) === "WAIT_FOR_SIGNATURE"
+          ? enqueueRequestAwaitingSignature(signatureRequest)
+          : TE.right(signatureRequest)
+      ),
```
gunzip:
```suggestion
      TE.filterOrElse(flow(status,  !S.Eq.equals("WAIT_FOR_SIGNATURE")),
          enqueueRequestAwaitingSignature
      ),
```

src/app/use-cases/validate-document.ts
```diff
@@ -68,5 +71,10 @@ export const makeValidateDocument =
...
       TE.map(() => void 0)
```
gunzip:
```suggestion
```

src/app/use-cases/validate-document.ts
```diff
@@ -68,5 +71,10 @@ export const makeValidateDocument =
         )
```
gunzip, line 71: may you add a comment on what the method 'makeValidateDocument' is supposed to do? explaining why we're putting the message into the queue.

src/infra/azure/storage/queue/signature-request.ts
gunzip: I see. It looks like the only thing this method add to `enqueueMessage` is the `JSON.stringify `part, probably you don't need this method at all as is. 

What you really want is to define a new type for each of the different kind of messages and an encoder (JSON.stringify) for them. In this case, it would be a method that takes a `signatureRequest` object and converts it to the structure the queue message is expecting. This would enable the consumer and producer to share this type to encode / decode it.

Something like: `SignatureRequestMessage.encode(signatureRequest)`

src/ui/api-models/Clause.ts
```diff
@@ -10,7 +10,7 @@ import * as t from "io-ts";
 const ClauseR = t.interface({
   title: t.string,
 
-  required: t.boolean,
+  required: t.boolean
```
gunzip: I'm unsure if we want to commit generated files. cc @lucacavallaro what about it?

src/app/use-cases/validate-document.ts
grausof: `enqueueRequestAwaitingSignature` return a `TaskEither` and I can't use a `filterOrElse`, furthermore `S.Eq.equals` requires you to pass the second parameter without a pipe. 
I could change like that:

```
TE.chain((signatureRequest) =>
        S.Eq.equals("WAIT_FOR_SIGNATURE", status(signatureRequest))
          ? enqueueRequestAwaitingSignature(signatureRequest)
          : TE.right(signatureRequest)
      )
```

gunzip: sorry, no it's ok as is

src/ui/api-models/Clause.ts
grausof: for me it is better not to commit them.

gunzip: deal, put them into some `generated` folder then add them to .gitingore

src/app/use-cases/validate-document.ts
grausof: I note that `TE.map(() => void 0)` is necessary for the handler-kit to work properly. CC @lucacavallaro

src/ui/api-models/Clause.ts
grausof: in this way we can delete the `ui` folder?

gunzip: Yes if it just contains generated files. 

Consider to generate the types for queue message (from a different yaml) as well. In other words, since other processes (external to this repo) may read the queue messages, better to threat them as API types.

grausof: ok I move everything to a `generated` folder. For now, the types for queue messages are used only by this repository. However as soon as possible we will create a dedicated yaml.

src/infra/services/service.ts
grausof: Done!

local.settings.json.example
```diff
@@ -0,0 +1,15 @@
+{
```
gunzip, line 1: can we use a .env file instead? (so we can add commments)

src/infra/IOApiClient/client.ts
```diff
@@ -0,0 +1,38 @@
...
+      fetchApi: toFetch(
+        setFetchTimeout(
+          config.ioapi.serviceRequestTimeout as Millisecond,
+          abortableFetch
+        )
+      ) as unknown as typeof fetch,
+      withDefaults: (op) => (params) =>
+        op({
+          ...params,
+          // please refer to source api spec for actual header mapping
```
gunzip: is this comment relevant?

grausof: We can take it off!

local.settings.json.example
grausof: `local.settings.json` is the default file used by azure function just for local usage but in production, these values have to be populated in Application settings. I think it is not possible to rename it!

lucacavallaro: maybe we can add a `local.settings.json.example` file

grausof: @lucacavallaro already added in PR!

lucacavallaro (approved):
Bravo Francesco 👏

PR #17, opened 2022-08-02
Refactor folder structure [SFEQS-922]

grausof:
this PR improves the organization of folders within the project


CreateProduct/function.json
```diff
@@ -14,5 +14,5 @@
       "name": "$return"
```
gunzip, line 14: it looks like you've deleted the files (instead of moving them)

gunzip (changes requested):

gunzip (approved):

CreateProduct/function.json
grausof: check now!

PR #20, opened 2022-08-29
[SFEQS-933] Add "qrCodeUrl" property to signature request

lucacavallaro:
#### List of Changes
1. Added the `qrCodeUrl` property on `SignatureRequest`

#### Types of changes

#### Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/openapi-issuer.yaml
+++ b/openapi-issuer.yaml
@@ -312,6 +312,7 @@ components:
     QrCodeUrl:
       type: string
       format: url
+      minLength: 1
 
     RequestSignatureBody:
       type: object
--- a/openapi-issuer.yaml
+++ b/openapi-issuer.yaml
@@ -308,6 +308,11 @@ components:
       required:
         - id
         - documents
+
+    QrCodeUrl:
+      type: string
+      format: url
+
     RequestSignatureBody:
       type: object
       properties:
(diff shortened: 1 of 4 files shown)


src/app/use-cases/request-signature.ts
```diff
@@ -62,6 +62,10 @@ export const makeRequestSignature =
           productId: payload.productId,
           signerId: signer.id,
           documents,
+          // WARNING! this is just a placeholder
+          // TODO: replace the static QR-code with a dynamic one
+          qrCodeUrl:
```
gunzip: can we put this into some constant named "mockSomething" (too have a recognizable pattern that clarifies this is a stub)? moreover I'd rather link something that's inside this repository.

openapi-issuer.yaml
```diff
@@ -308,6 +308,11 @@ components:
       required:
         - id
         - documents
+
+    QrCodeUrl:
+      type: string
+      format: url
```
gunzip, line 314: minLength: 1 ?

src/app/use-cases/request-signature.ts
lucacavallaro: sure

PR #21, opened 2022-08-30
Rename queue for message [SFEQS-957]

grausof:
This PR rename the queue for message from "waiting_for_signature" to "waiting_for_message"!

### DIFF
--- a/SendSignatureRequestMessage/function.json
+++ b/SendSignatureRequestMessage/function.json
@@ -4,7 +4,7 @@
       "name": "sendSignatureRequestMessage",
       "type": "queueTrigger",
       "direction": "in",
-      "queueName": "waiting-for-signature",
+      "queueName": "waiting-for-message",
       "connection": "AzureWebJobsStorage"
     }
   ],
--- a/SendSignatureRequestMessage/function.json
+++ b/SendSignatureRequestMessage/function.json
@@ -4,7 +4,7 @@
       "name": "sendSignatureRequestMessage",
       "type": "queueTrigger",
       "direction": "in",
-      "queueName": "waiting-for-message",
+      "queueName": "%WaitingMessageQueueName%",
       "connection": "AzureWebJobsStorage"
     }
   ],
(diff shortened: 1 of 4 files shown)


SendSignatureRequestMessage/function.json
```diff
@@ -4,7 +4,7 @@
       "name": "sendSignatureRequestMessage",
       "type": "queueTrigger",
       "direction": "in",
-      "queueName": "waiting-for-signature",
+      "queueName": "waiting-for-message",
```
gunzip: you may use the environment variable: `%QUEUE_STORAGE_NAME_FOR_WAITING_MESSAGE%`

grausof: you are right, thanks!

grausof: it is possible to directly use the environment variable with the syntax `%%`, so I entered `%WaitingMessageQueueName%`!

SendSignatureRequestMessage/function.json
```diff
@@ -4,7 +4,7 @@
       "name": "sendSignatureRequestMessage",
       "type": "queueTrigger",
       "direction": "in",
-      "queueName": "waiting-for-signature",
+      "queueName": "%WaitingMessageQueueName%",
```
gunzip, line 7: I think it's %QUEUE_STORAGE_NAME_FOR_WAITING_MESSAGE% (local settings isnt' read in production)

grausof: `QUEUE_STORAGE_NAME_FOR_WAITING_MESSAGE` is only a placeholder for local settings. The real env name is `WaitingMessageQueueName`. https://github.com/pagopa/io-functions-sign/blob/187b304741b898e0589b54dfc3fb777b41c64051/src/infra/azure/storage/webjobs-config.ts#L25

gunzip: Ok!

gunzip (approved):

PR #22, opened 2022-09-12
Mark signature request as ready [SFEQS-950]

grausof:
This PR adds a new method (PATCH) to change the status of the signature request from WAIT_FOR_ISSUER to READY. The new states introduced are:
 - DRAFT: The signature request has been created but not all documents have been uploaded yet.
 - WAIT_FOR_ISSUER: Documents have been uploaded but the ISSUER has not yet marked the signature request as READY.
 - READY: The signature request is ready for processing. In this phase the documents will be analyzed and prepared for sending to the USER.
 - WAIT_FOR_SIGNATURE: The signature request has been sent to the user and is awaiting signature
 - SIGNED: The signature request was successfully signed.

#### List of Changes

- removed the function that calculated the status. Now the state is saved on the DB
- added a new openapi specification

#### Types of changes

- [X] New feature (non-breaking change which adds functionality)

#### Checklist:
- [X] I have updated the documentation accordingly.


openapi-issuer.yaml
```diff
@@ -327,6 +352,15 @@ components:
         - productId
         - fiscalCode
 
+    PatchSignatureStatusBody:
+      type: object
+      properties:
+        status:
+          type: string
```
lucacavallaro: here you could use an `enum` instead of a string

src/app/use-cases/request-signature.ts
```diff
@@ -71,3 +81,54 @@ export const makeRequestSignature =
...
+  (
+    payload: RequestSignatureStatusPayload
+  ): TE.TaskEither<Error, SignatureRequest> =>
+    pipe(
+      sequenceS(TE.ApplySeq)({
+        signatureRequestStatus: pipe(
+          TE.of(payload.signatureRequestStatus),
+          TE.filterOrElse(
+            (_) => _ === "READY",
+            constant(new InvalidEntityError("Only READY status is allowed!"))
```
lucacavallaro: you should use `ValidationError` (from handler-kit) instead `InvalidEntityError`

try to move this check to `io-ts` :)

src/app/use-cases/request-signature.ts
```diff
@@ -71,3 +81,54 @@ export const makeRequestSignature =
...
+            payload.subscriptionId
+          ),
+          TE.chainEitherKW(
+            flow(
+              E.fromOption(
+                () =>
+                  new EntityNotFoundError("Error getting the Signature Request")
+              ),
+              E.filterOrElse(
+                (_) => _.status === "WAIT_FOR_ISSUER",
```
lucacavallaro: `_` should be used only as placeholder for "unused variable"

lucacavallaro (changes requested):

openapi-issuer.yaml
```diff
@@ -108,6 +108,31 @@ paths:
           $ref: "#/components/responses/NotFound"
         default:
           $ref: "#/components/responses/Unexpected"
+    patch:
+      operationId: patchSignatureRequestStatus
+      summary: Change status of a signature request
```
gunzip:
```suggestion
      summary: Change the status of a signature request
```

openapi-issuer.yaml
gunzip: should be an enum see 

https://github.com/pagopa/io-functions-services/blob/master/openapi/index.yaml#L802

src/app/use-cases/request-signature.ts
grausof: Thanks, I will try to move it to `io-ts`. 
But `InvalidEntityError` requires `violations: t.Errors`, where should I get it from? Also from a quick search, `ValidationError` is never used. 😬

src/app/use-cases/request-signature.ts
```diff
@@ -71,3 +81,54 @@ export const makeRequestSignature =
...
+  (upsertSignatureRequest: UpsertSignatureRequest) =>
+  (
+    payload: RequestSignatureStatusPayload
+  ): TE.TaskEither<Error, SignatureRequest> =>
+    pipe(
+      sequenceS(TE.ApplySeq)({
+        signatureRequestStatus: pipe(
+          TE.of(payload.signatureRequestStatus),
+          TE.filterOrElse(
+            (_) => _ === "READY",
```
gunzip: limit usage of '_', use meaningful names instead.

src/app/use-cases/validate-document.ts
```diff
@@ -67,13 +65,17 @@ export const makeValidateDocument =
         pipe(
           request,
           addUrlToDocument(payload.documentId, payload.documentUrl),
-          E.map((documents) => ({ ...request, documents }))
+          E.map((documents) => ({ ...request, documents })),
+          E.map((request) => ({
+            ...request,
+            status:
+              request.documents.every(
+                (document) => document.url && document.url.length > 0
```
gunzip:
```suggestion
                (document) => NonEmptyString.is(document)
```

src/app/use-cases/request-signature.ts
lucacavallaro: Actually you used `ValidationError` in this PR, in fact it is built by the `validate` function of `handler-kit`

src/app/use-cases/validate-document.ts
```diff
@@ -67,13 +65,17 @@ export const makeValidateDocument =
...
+              ) && request.status === "DRAFT"
```
gunzip: you must isolate this state machine into a switch statement in its own method, somehing like:

`const next = (prev: Status) => Status`

src/app/use-cases/request-signature.ts
lucacavallaro: After creating a `io-ts` decoder that checks the `status`, you can receive a `ValidationError` by passing it to the `validate` function.

src/signature-request/signature-request.ts
```diff
@@ -14,9 +14,20 @@ import { ProductId } from "./product";
...
 
+/*
+ * Signature request status meaning:
+ * - DRAFT: The signature request has been created but not all documents have been uploaded yet.
+ * - WAIT_FOR_ISSUER: Sll documents have been uploaded but the ISSUER has not yet marked the signature request as READY.
+ * - READY: The signature request is ready for processing. In this phase the documents will be analyzed and prepared for sending to the USER.
+ * - WAIT_FOR_SIGNATURE: The signature request has been sent to the user and is awaiting signature
+ * - SIGNED: The signature request was successfully signed.
+ */
 export const SignatureRequestStatus = t.keyof({
```
gunzip, line 26: I guess this may be generated by the API specs if you use enums

gunzip (changes requested):
lacks unit tests

openapi-issuer.yaml
grausof: Already reported and fixed!

src/app/use-cases/request-signature.ts
grausof: Already reported and fixed!

src/signature-request/signature-request.ts
lucacavallaro: @gunzip yep but this structure is used by the "domain layer"

gunzip: ok

grausof: Yes, thank you! I added the self-generated enum!

src/app/use-cases/request-signature.ts
grausof: OK thank you. Now it is clearer!

gunzip: t.Errors is the left result when calling `decode()`

src/app/use-cases/validate-document.ts
grausof: In this case the change of status is also linked to the presence of the documents. A possible implementation could be:
`const status = (request: SignatureRequest) => Status`. 
Otherwise I didn't understand.

gunzip: I mean that all constraints on status update (from the previous state) must be isolated. What we want to avoid is arbitrary invalid status changes (ie. SIGNED -> DRAFT).

If the changes depends on the whole signature request (and only on that) you may use 

`const nextStatus = (request: SignatureRequest) => Status`

otherwise `const next = (prev: Status) => Status` is enough.

gunzip:
why don't we have coverage information?

openapi-issuer.yaml
```diff
@@ -108,6 +108,31 @@ paths:
           $ref: "#/components/responses/NotFound"
         default:
           $ref: "#/components/responses/Unexpected"
+    patch:
+      operationId: patchSignatureRequestStatus
```
gunzip: Try not to name things leaking implementation details that are subject to changes ('patch').
```suggestion
      operationId: setSignatureRequestStatus
```

openapi-issuer.yaml
```diff
@@ -108,6 +108,31 @@ paths:
...
+      summary: Change the status of a signature request
+      parameters:
+        - in: path
+          name: signatureRequestId
+          schema:
+            type: string
+            format: ulid
```
gunzip: is this format enfoced somewhat into the io-ts type?

openapi-issuer.yaml
```diff
@@ -108,6 +108,31 @@ paths:
...
+          required: true
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/PatchSignatureStatusBody"
+      responses:
+        "204":
+          description: The signature status was changed!
```
gunzip:
```suggestion
          description: Signature status successfully changed.
```

grausof:
> why don't we have coverage information?

@gunzip Coverage is generated but not published by the pipeline! We can add it as a task!

openapi-issuer.yaml
```diff
@@ -108,6 +108,31 @@ paths:
           $ref: "#/components/responses/NotFound"
         default:
           $ref: "#/components/responses/Unexpected"
+    patch:
```
gunzip: Thinking about this, I'd rather use another endpoint` /signature-requests/{id}/status`
and use a `PUT` accepting only the new status as payload (not the whole request).

that's because you're not patching the whole object here, but you're limiting the options to just one field.

openapi-issuer.yaml
gunzip: I'd return the new status

src/app/use-cases/request-signature.ts
```diff
@@ -28,6 +36,27 @@ export type RequestSignaturePayload = {
...
+  SignatureRequestStatus,
+  SignatureRequestStatus,
+  unknown
+>(
+  "status",
+  isReadyStatus,
+  (input, context) =>
+    isReadyStatus(input) ? t.success(input) : t.failure(input, context),
+  t.identity
+);
```
gunzip: this looks like overkill and there's some type mangling. an inline comparison is enough

src/app/use-cases/request-signature.ts
```diff
@@ -71,3 +103,51 @@ export const makeRequestSignature =
...
+            flow(
+              E.fromOption(
+                () =>
+                  new EntityNotFoundError("Error getting the Signature Request")
+              ),
+              E.filterOrElse(
+                (request) => request.status === "WAIT_FOR_ISSUER",
+                constant(
+                  new InvalidEntityError(
+                    "The status can only be changed if the signature request is still in WAIT_FOR_ISSUER!"
```
gunzip: try to isolate this logic into `nextStatus`

src/app/use-cases/validate-document.ts
```diff
@@ -45,12 +45,17 @@ export type ValidateDocumentPayload = GetDocumentPayload & {
   documentUrl: string;
 };
 
+const nextStatus = (request: SignatureRequest) =>
+  request.documents.every((document) => NonEmptyString.is(document)) &&
+  request.status === "DRAFT"
+    ? "WAIT_FOR_ISSUER"
+    : request.status;
```
gunzip: you may choose to return an option and none in case of failure (or an either in case you want to store a meaningful error string)

gunzip (changes requested):

openapi-issuer.yaml
grausof: ok, maybe it is better to return the entire signature request

openapi-issuer.yaml
grausof: I was looking at the [Zalando guidelines](https://opensource.zalando.com/restful-api-guidelines/#patch), and i used PATCH just to change only part of the resource and not the whole request (`PUT`).

Do you think maybe it is better to add `PUT` on  `/signature-requests/{id}/status` for a better comprehension?

gunzip: Using `PATCH` is a feasible option if caller the can change anything about the resources (moreover Zalando guidelines suggest to use a specific format to do that). The ratio is that `PATCH` is preferred when you only want to send the "patch document" instead of the whole object (that may be big), so you're sure that you're only affecting the sent fields (but this usually doesn't limit you a subset and this limit is not communicated by the semantic).

In this case the caller can change only the status field so I'd rather go for PUT with a /status URL.

@lucacavallaro what about this?

grausof: I agree!

src/app/use-cases/request-signature.ts
grausof: according to https://github.com/pagopa/io-functions-sign/pull/22#discussion_r968613274 , to use the validate function I must necessarily create a codec!

gunzip: you can use `t.literal("READY")` then

openapi-issuer.yaml
lucacavallaro: Both are excellent options, so the one you like best is fine 👍 

If we want to keep the `PATCH` verb and make it more future-proof, we can use the standard `JSON PATCH` https://www.rfc-editor.org/rfc/rfc6902

openapi-issuer.yaml
grausof: @lucacavallaro

src/app/use-cases/request-signature.ts
grausof: `nextStatus` is meant to calculate a new state from a previous one, in this case we just need to check if the signature request is in the" WAIT_FOR_ISSUER "state. 

However do you think it is more logical to create such a function like:
`const nextStatus: (request: SignatureRequest) => (newStatus: SignatureRequestStatus) => O.Option<SignatureRequestStatus>` which checks if the new state is admitted and returns it?

gunzip: ok.

probably, you choose.

src/app/use-cases/validate-document.ts
```diff
@@ -45,12 +46,24 @@ export type ValidateDocumentPayload = GetDocumentPayload & {
   documentUrl: string;
 };
 
+const nextStatus = (
```
gunzip: this method should handle all cases (not only from DRAFT to WAIT_FOR_ISSUER),
you probably want to use a switch statement here.

src/app/use-cases/request-signature.ts
```diff
@@ -71,3 +88,38 @@ export const makeRequestSignature =
...
+  ) =>
+  (
+    payload: RequestSignatureStatusPayload
+  ): TE.TaskEither<Error, SignatureRequest> =>
+    pipe(
+      payload.signatureRequestStatus,
+      validate(t.literal("READY"), "Only READY status is allowed!"),
+      TE.fromEither,
+      TE.chain((_) =>
+        pipe(
```
gunzip: can we avoid the nested pipe here?

src/app/use-cases/status-signature-request.ts
```diff
@@ -0,0 +1,123 @@
+import * as E from "fp-ts/Either";
+import * as t from "io-ts";
+
+import { pipe } from "fp-ts/function";
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+
+import { ProblemDetail } from "@pagopa/handler-kit/lib/problem-detail";
+import { SignatureRequest } from "../../signature-request/signature-request";
+
+export class ActionNotAllowedError extends Error implements ProblemDetail {
```
gunzip: I guess this should go into some errors.ts module?

src/app/use-cases/status-signature-request.ts
```diff
@@ -0,0 +1,123 @@
...
+      case "WAIT_FOR_SIGNATURE":
+        return pipe(action, whenWaitForSignature(request));
+      case "SIGNED":
+        return E.left(
+          new ActionNotAllowedError(
+            "This action is not allowed when signature request is in already SIGNED!"
+          )
+        );
+      default:
+        return E.left(new ActionNotAllowedError("Status not valid!"));
```
gunzip:
```suggestion
        return E.left(new ActionNotAllowedError("Invalid status."));
```

src/app/use-cases/status-signature-request.ts
```diff
@@ -0,0 +1,123 @@
...
+      case "WAIT_FOR_ISSUER":
+        return pipe(action, whenWaitForIssuer(request));
+      case "READY":
+        return pipe(action, whenReady(request));
+      case "WAIT_FOR_SIGNATURE":
+        return pipe(action, whenWaitForSignature(request));
+      case "SIGNED":
+        return E.left(
+          new ActionNotAllowedError(
+            "This action is not allowed when signature request is in already SIGNED!"
```
gunzip: I suggest to use quill to correct spelling https://quillbot.com/

src/app/use-cases/status-signature-request.ts
```diff
@@ -0,0 +1,123 @@
...
+        return E.left(new ActionNotAllowedError("Status not valid!"));
+    }
+  };
+
+const whenDraft =
+  (request: SignatureRequest) =>
+  (
+    action: SignatureRequestAction
+  ): E.Either<ActionNotAllowedError, SignatureRequest> =>
+    action === "UPLOAD_DOCUMENT"
```
gunzip: I'd use a switch here as well

src/app/use-cases/status-signature-request.ts
```diff
@@ -0,0 +1,123 @@
...
+        );
+    }
+  };
+
+const whenWaitForSignature =
+  (request: SignatureRequest) =>
+  (
+    action: SignatureRequestAction
+  ): E.Either<ActionNotAllowedError, SignatureRequest> =>
+    action === "MARK_AS_SIGNED"
```
gunzip: ditto

src/app/use-cases/status-signature-request.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export const SignatureRequestAction = t.keyof({
+  UPLOAD_DOCUMENT: null,
+  MARK_AS_READY: null,
+  VALIDATE_DOCUMENT: null,
+  MARK_AS_SIGNED: null,
+});
+export type SignatureRequestAction = t.TypeOf<typeof SignatureRequestAction>;
+
+export const nextStatus =
```
gunzip: very good!

src/app/use-cases/status-signature-request.ts
gunzip: It reminds to me this https://xstate.js.org/docs/ (or redux). Maybe we can consider to use xstate for this kind of state machine (but you have to create a singleton state machine then pass it to the methods that need it)

gunzip: Another use of xstate is logging, see `onTransition` here: https://xstate.js.org/docs/

gunzip: @lucacavallaro what do you think about it?

src/app/use-cases/status-signature-request.ts
gunzip: "This operation is prohibited if the signature request has already been signed."

src/app/use-cases/status-signature-request.ts
grausof: lint suggests: `switch statements should have at least 3 case clauseses` for this reason I used an inline check.

...
For this type of implementation I took a cue from xstate, but it seemed excessive to use the entire library.

gunzip: I agree more with sonarcloud:
image

so I suggest to selectively disable linting

grausof: ok, I take the control off!

src/app/use-cases/request-signature.ts
```diff
@@ -71,3 +88,32 @@ export const makeRequestSignature =
...
+      ),
+      TE.chainEitherKW(
+        flow(
+          E.fromOption(
+            () => new EntityNotFoundError("Error getting the Signature Request")
+          ),
+          E.chainW((request) => pipe("MARK_AS_READY", nextStatus(request)))
+        )
+      ),
+      TE.chain(upsertSignatureRequest)
```
gunzip:
```suggestion
      TE.chain(() =>
        getSignatureRequest(payload.signatureRequestId, payload.subscriptionId)
      ),
      TE.chainW(
        TE.fromOption(
          () => new EntityNotFoundError("Error getting the Signature Request")
        )
      ),
      TE.chainEitherKW((request) => pipe("MARK_AS_READY", nextStatus(request))),
      TE.chain(upsertSignatureRequest)
```

gunzip: If you invert the arguments in `nextStatus` you can get:

`TE.chainEitherKW(nextStatus("MARK_AS_READY"))`

for the last line

gunzip: Try to avoid nested pipelines / flow

grausof: right, I agree!

src/app/use-cases/__test__/status-signature-request.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+        action: "VALIDATE_DOCUMENT",
+      },
+      expected: true,
+    },
+  ])("should be valid ($#)", ({ payload, expected }) => {
+    const makeRequest = pipe(
+      payload.request as SignatureRequest,
+      nextStatus(payload.action as SignatureRequestAction),
+      E.mapLeft((e) => {
+        expect(e).toBeInstanceOf(ActionNotAllowedError);
```
gunzip: this expectation won't run if e is right. You have two alternatives:

1. write an expectation before this one which assert that e is left
1. declare the number of expectation at the begin of the test using `expect.assertions(<number>)`

src/app/use-cases/__test__/request-signature.spec.ts
```diff
@@ -62,3 +67,42 @@ describe("MakeRequestSignatureList", () => {
...
+      updateStatusRequestSignature(
+        mockUpsertSignatureRequest,
+        mockGetSignatureRequest
+      )
+    )();
+    return updateRequest.then((data) => {
+      pipe(
+        data,
+        E.mapLeft((e) => {
+          expect(e).toBeInstanceOf(ValidationError);
```
gunzip: see other comment

src/app/use-cases/status-signature-request.ts
```diff
@@ -0,0 +1,122 @@
+/* eslint-disable sonarjs/no-small-switch */
```
gunzip: you usually want to put this kind of comments near the code selectively disablling "next line" (otherwise when you refactor your code and  the switches are deleted (for example) these comments will remain even if unused.

src/app/use-cases/status-signature-request.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export const SignatureRequestAction = t.keyof({
+  UPLOAD_DOCUMENT: null,
+  MARK_AS_READY: null,
+  VALIDATE_DOCUMENT: null,
+  MARK_AS_SIGNED: null,
+});
+export type SignatureRequestAction = t.TypeOf<typeof SignatureRequestAction>;
+
+export const nextStatus =
```
gunzip:
```suggestion
export const setNextStatusAs =
```

gunzip: you can use fn + f2 to refactor names

gunzip (changes requested):
almost there

src/app/use-cases/__test__/status-signature-request.spec.ts
grausof: but the expectation is always only one. If e is right I don't have to check the type of error....

src/app/use-cases/status-signature-request.ts
grausof: 🥷

src/app/use-cases/__test__/status-signature-request.spec.ts
gunzip: All the expectation in a test are supposed to run. if not you probably want to restructure you're tests or
 use simple if statements for this kind of tests using `E.isRight / E.isLeft` to make it more clear

grausof: ok, I cleared the error type check. Maybe that can be done in a separate test

openapi-issuer.yaml
```diff
@@ -327,14 +357,33 @@ components:
         - productId
         - fiscalCode
 
+    PutSignatureStatusBody:
+      type: object
+      properties:
+        status:
+          $ref: "#/components/schemas/StatusSignatureRequest"
```
gunzip:
```suggestion
          $ref: "#/components/schemas/SignatureRequestStatus"
```

openapi-issuer.yaml
```diff
@@ -327,14 +357,33 @@ components:
...
+      required:
+        - status
+
+    StatusSignatureRequest:
+      type: string
+      enum:
```
gunzip:
```suggestion
      x-extensible-enum:
```

see https://opensource.zalando.com/restful-api-guidelines/#112

src/app/use-cases/validate-document.ts
```diff
@@ -67,13 +65,9 @@ export const makeValidateDocument =
...
+          E.chain(dispatch("UPLOAD_DOCUMENT"))
         )
       ),
-      TE.chain(upsertSignatureRequest),
-      TE.chain((signatureRequest) =>
-        status(signatureRequest) === "WAIT_FOR_SIGNATURE"
-          ? enqueueRequestAwaitingSignature(signatureRequest)
-          : TE.right(signatureRequest)
-      )
+      TE.chain(upsertSignatureRequest)
```
gunzip, line 72: it looks like `enqueueRequestAwaitingSignature` gets lost, is this on purpose?

src/infra/azure/functions/put-signature-request-status.ts
```diff
@@ -0,0 +1,89 @@
...
+    req,
+    body(PutSignatureStatusBody),
+    E.map((body) => body.status)
+  );
+
+export const extractPatchSignatureRequestPayload: RE.ReaderEither<
+  HttpRequest,
+  Error,
+  {
+    subscriptionId: Subscription["id"];
```
gunzip, line 49: it looks like you are using the internal model types. since these vaues comes from the API payload, I guess these types here should be imported from api-models instead, isn't it @lucacavallaro ?

src/infra/azure/functions/put-signature-request-status.ts
```diff
@@ -0,0 +1,89 @@
...
+  upsertSignatureRequest,
+  getSignatureRequest
+);
+
+export const requireSignatureRequestId: (
+  req: HttpRequest
+) => E.Either<Error, SignatureRequest["id"]> = flow(
+  path("signatureRequestId"),
+  E.fromOption(() => new Error("Missing signatureRequestId in path")),
+  E.chainW(validate(SignatureRequestId, "Invalid signatureRequestId in path"))
```
gunzip: I'm not sure we should use the interal model type instead of the API (generated) one. See the following comment.

src/app/use-cases/validate-document.ts
grausof: Yes, `enqueueRequestAwaitingSignature` was used to send the message right away. Later this part will still be reviewed with another PR to send the documents for validation.

src/infra/azure/functions/put-signature-request-status.ts
lucacavallaro: nope, this function is usually the one we use to convert presentation layer types to domain types (which are the ones then used by the use-case)

so, in my opinion, it's okay to use these types of domains here

lucacavallaro: no, at the moment we do not perform ULID runtime validation

(I'll add it on Jira)

openapi-issuer.yaml
lucacavallaro: nope, right now we don't validate ULID at runtime

gunzip (approved):

lucacavallaro (approved):

PR #25, opened 2022-10-05
Add status on DOCUMENT [SFEQS-1011]

grausof:
This PR add status on DOCUMENT:
- WAIT_FOR_UPLOAD: The document has not been uploaded yet.
- VALIDATION_IN_PROGRESS: Validation of the uploaded document in progress.
- READY: The document has passed validation and is ready.
- VALIDATION_ERROR: An error occurred while validating the document.

The status of the SIGNATURE REQUEST have been changed to:
- DRAFT: The signature request has been created but not all documents have been uploaded yet.
- READY: The signature request is ready for processing. In this phase the documents will be analyzed and prepared for sending to the USER.
- WAIT_FOR_SIGNATURE: The signature request has been sent to the user and is awaiting signature
- SIGNED: The signature request was successfully signed.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/app/use-cases/validate-document.ts
+++ b/src/app/use-cases/validate-document.ts
@@ -1,9 +1,9 @@
 import * as TE from "fp-ts/TaskEither";
 import * as O from "fp-ts/Option";
 import * as E from "fp-ts/Either";
+import * as A from "fp-ts/Array";
 import { identity, pipe } from "fp-ts/function";
 
-import { findIndex, modifyAt } from "fp-ts/Array";
 import { validate } from "@pagopa/handler-kit/lib/validation";
 import {
   GetSignatureRequest,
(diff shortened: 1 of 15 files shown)


src/app/use-cases/validate-document.ts
```diff
@@ -18,25 +18,51 @@ import {
...
+        pipe(
+          request.documents,
+          A.lookup(index),
+          E.fromOption(() => documentNotFoundError),
+          E.chain(dispatchOnDocument(action)),
+          E.chain((document) =>
+            pipe(
+              request.documents,
+              A.modifyAt(index, () => document),
+              E.fromOption(() => documentNotFoundError)
```
gunzip, line 61: you don't need this line since it's already checked

src/app/use-cases/validate-document.ts
```diff
@@ -18,25 +18,51 @@ import {
...
-          modifyAt(index, (document) => ({
+          A.modifyAt(index, (document) => ({
             ...document,
             url,
           }))
         )
       ),
-      E.fromOption(() => new EntityNotFoundError("Document not found")),
+      E.fromOption(() => documentNotFoundError),
+      E.chainW(validate(DocumentList, "Unable to validate document list"))
```
gunzip, line 41: why do you want to revalidate request.documents which is already a document list?

src/app/use-cases/validate-document.ts
```diff
@@ -18,25 +18,51 @@ import {
...
+          E.chain((document) =>
+            pipe(
+              request.documents,
+              A.modifyAt(index, () => document),
+              E.fromOption(() => documentNotFoundError)
+            )
+          )
+        )
+      ),
       E.chainW(validate(DocumentList, "Unable to validate document list"))
```
gunzip, line 66: ditto

src/infra/azure/functions/validate-document-upload.ts
```diff
@@ -52,6 +55,11 @@ const validateDocument = makeValidateDocument(
   isDocumentUploadedToBlobStorage
 );
 
+const updateDocumentStatus = makeChangeDocumentStatus(
+  getSignatureRequest,
```
gunzip, line 59: It's not clear to me what this method and this one:

```
const validateDocument = makeValidateDocument(
  getSignatureRequest,
  upsertSignatureRequest,
  isDocumentUploadedToBlobStorage
);
```

are supposed to do. It looks like they just get the signature request and save the same content into the db.

src/signature-request/product.ts
```diff
@@ -35,11 +35,14 @@ export type GetProduct = (
...
-  })),
+  map(
+    (metadata): Document => ({
+      id: id(),
+      status: "WAIT_FOR_UPLOAD",
+      ...metadata,
+      ...timestamps(),
+    })
+  ),
   DocumentList.decode,
```
gunzip, line 46: do we need this decode here?

src/app/use-cases/validate-document.ts
grausof: `A.modifyAt(index, () => document)` change an element in the array and return an option. So the check on the option must be done anyway

src/app/use-cases/validate-document.ts
grausof: If it is not revalidated it fails to verify the branded types. 
`Property '[_brand]' is missing in type`

src/infra/azure/functions/validate-document-upload.ts
grausof: `makeChangeDocumentStatus` only changes the state of the document on the DB based on the current state, while `makeValidateDocument` takes care of validating the document and then adding the url to the DB. These functions have been separated because before starting the validation with `makeValidateDocument` it is necessary to change the status of the document to `WAIT_FOR_VALIDATION`. Having everything in a single function, the status of the document would change only at the end of the whole validation process.

src/signature-request/product.ts
grausof: ditto

src/infra/azure/functions/validate-document-upload.ts
gunzip: I'm wondering how since by the code it looks they do the upsert just after getting the object from the database (isn't it te same object that gets written?)

gunzip (approved):

PR #26, opened 2022-10-10
Add user OpenApi [SFEQS-726]

grausof:
This PR begins to describe the openapi specification for the user. It introduces a new `Signature` entity that keeps track of all the signatures affixed to the various fields and integration with the QTSP

### DIFF
--- a/docs/openapi-user.yaml
+++ b/docs/openapi-user.yaml
@@ -311,6 +311,14 @@ components:
       format: base64url
       description: SPID SAML assertion encoded in base64
 
+    ClausesSignature:
+      type: object
+      properties:
+        signatureFieldId:
+          type: string
+        accepted:
+          type: boolean
+
     DocumentSignature:
       type: object
       properties:
@@ -321,9 +329,14 @@ components:
           type: string
           format: base64url
           description: Signature of document digest concatenated with document clauses
+        clauses:
+          type: array
+          items:
+            $ref: "#/components/schemas/ClausesSignature"
       required:
         - documentId
         - signature
+        - clauses
 
     QtspClausesMetadata:
       type: object
--- a/docs/openapi-user.yaml
+++ b/docs/openapi-user.yaml
@@ -1,170 +1,406 @@
 openapi: 3.0.3
-
 info:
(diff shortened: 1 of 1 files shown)


gunzip:
lgtm (I don't have enough context to enter into details)

docs/openapi-user.yaml
```diff
@@ -1,170 +1,419 @@
 openapi: 3.0.3
-
 info:
-  title: Firma con IO User APIs
-  description: Any APIs called by IO App
-  version: 1.0.0
-
+  title: IO Sign - User API
```
hevelius, line 3: What do you think about using "Firma con IO - User API"?

docs/openapi-user.yaml
```diff
@@ -1,170 +1,419 @@
...
+      format: base64url
+      description: SPID SAML assertion encoded in base64
+
+    ClausesSignature:
+      type: object
+      properties:
+        signatureFieldId:
+          type: string
+        accepted:
+          type: boolean
```
hevelius, line 320: The clauses must not also have a required field boolean? Should the document clauses in the UI not be mandatory or optional?

docs/openapi-user.yaml
grausof: I agree. Thank you!

grausof: Other references also have the name "io-sign" including the repository. For now I would leave it like this!

docs/openapi-user.yaml
grausof: Yes, to show them you have to retrieve them from the `signature-request`. In this case it is the POST of the accepted clauses where it is not necessary to know if they are mandatory or not.

hevelius: Great! Thanks!

docs/openapi-user.yaml
hevelius: In the context of IO the name "Firma con IO" is used I thought it was better to refer to this :-D

hevelius (approved):
LGTM

lucacavallaro (approved):

PR #27, opened 2022-10-19
docs: add OpenApi for user [SFEQS-1106]

grausof:
This PR add a DRAFT of OpenApi for user and backend


docs/openapi-user.yaml
```diff
@@ -93,183 +109,50 @@ paths:
           $ref: "#/components/responses/Unexpected"
 
 components:
+  securitySchemes:
```
gunzip, line 177: must match this one: https://github.com/pagopa/io-functions-app/blob/master/openapi/index.yaml#L354

(header name is `X-Functions-Key`)

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,258 @@
+openapi: 3.0.3
+info:
+  title: IO Sign - Backend API
+  version: 0.0.1
+paths:
```
gunzip, line 10: add bearer security: https://github.com/pagopa/io-backend/blob/master/api_backend.yaml#L10

gunzip (changes requested):

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,258 @@
...
+            $ref: "#/components/schemas/DocumentSignature"
+          minItems: 1
+          uniqueItems: true
+        qtspClauses:
+          $ref: "#/components/schemas/QtspClausesMetadata"
+      required:
+        - signatureRequestId
+        - email
+        - publicKey
+        - spidSamlAssertion
```
hevelius: are they still needed?

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,258 @@
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/CreateSignatureBody"
+      responses:
+        "201":
+          description: Signature created
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SignatureDetailView"
```
hevelius: This call occurs when the user sends the signature request to the QTSP. Is it necessary that the response contain the SignatureDetailView? Or the status code is sufficient?

docs/openapi-backend.yaml
grausof: No. Thanks, I take these things off.

docs/openapi-backend.yaml
grausof: Status is sufficient!

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+    SignatureRequestId:
+      $ref: "https://raw.githubusercontent.com/pagopa/io-functions-sign/v1.0.0-RELEASE/openapi-issuer.yaml#/components/schemas/SignatureRequestId"
+
+    Email:
+      type: string
+      description: User's email.
+      format: EmailString
+      x-import: "@pagopa/ts-commons/lib/strings"
+      example: me@example.com
```
hevelius: should be removed?

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,207 @@
...
+      properties:
+        text:
+          type: string
+      required:
+        - text
+
+    PublicKey:
+      type: string
+      format: base64url
+      description: PEM public key encoded in base64
```
hevelius: ditto

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+    PublicKeyDigest:
+      type: string
+      format: base64url
+      description: Hash of PEM public key encoded in base64
+
+    SamlAssertion:
+      type: string
+      format: base64url
+      description: SPID SAML assertion encoded in base64
```
hevelius: ditto

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,185 @@
+openapi: 3.0.3
+info:
+  title: IO Sign - Backend API
+  version: 0.0.1
+paths:
```
hevelius, line 10: maybe here should be added a security section and a base path https://github.com/pagopa/io-backend/blob/aa7d7c64c71b582d87f229b15a6e26c59c80a76c/api_backend.yaml#L6-L11
What do you think?

grausof: Added. Check if the server is right

hevelius: the `basePath` should be `basePath: /api/v1/fci`

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,195 @@
...
+
+    ClausesSignature:
+      type: object
+      properties:
+        signatureFieldId:
+          type: string
+        accepted:
+          type: boolean
+
+    DocumentSignature:
```
hevelius, line 352: shouldn't have a `status` properties?

grausof: this will be changed, but for now let's leave it that way!

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,331 @@
...
+            x: { type: number, minimum: 0 }
+            y: { type: number, minimum: 0 }
+        page:
+          type: number
+          minimum: 0
+        size:
+          type: object
+          properties:
+            w: { type: number, minimum: 0 }
+            h: { type: number, minimum: 0 }
```
hevelius: Wouldn't it be better to use a one more definition for `Coords`?

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,331 @@
...
+    Clause:
+      type: object
+      properties:
+        title:
+          type: string
+        type:
+          type: string
+          enum: ["REQUIRED", "UNFAIR", "OPTIONAL"]
```
hevelius: Same thing for the enums. What do you think? @lucacavallaro @gunzip

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,331 @@
...
+      properties:
+        id:
+          $ref: "#/components/schemas/Id"
+        expires_at:
+          $ref: "#/components/schemas/Timestamp"
+        status:
+          type: string
+          enum:
+            - WAIT_FOR_SIGNATURE
+            - SIGNED
```
hevelius: ditto

docs/openapi-backend.yaml
grausof: I have reported the changes that were defined in #29

docs/openapi-backend.yaml
hevelius: Having a one more definition simplifies the use of imports.
`import { SignatureRequestStatusEnum } from "../../../../generated/definitions/fci/SignatureRequestStatus";` instead of
`import { StatusEnum as SignatureRequestStatusEnum } from "../../../../../../definitions/fci/SignatureRequestDetailView";`

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,331 @@
...
+      description: Hash of PEM public key encoded in base64
+
+    ClausesSignature:
+      properties:
+        type:
+          type: string
+          x-extensible-enum:
+            - COORDS
+            - UNIQUE_NAME
+          description: base64 of coords object or fieldName string
```
hevelius: ditto

hevelius (commented):
LGTM. As I wrote in the comments I would make the definitions more explicit. What do you think about it?

docs/openapi-backend.yaml
grausof: You are right. I have proceeded to separate the enums into a separate refs!

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,351 @@
...
+          $ref: "#/components/schemas/Id"
+        metadata:
+          $ref: "#/components/schemas/DocumentMetadata"
+        created_at:
+          $ref: "#/components/schemas/Timestamp"
+        updated_at:
+          $ref: "#/components/schemas/Timestamp"
+      required:
+        - id
+        - status
```
hevelius, line 149: where is `status`?

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,393 @@
+openapi: 3.0.3
+info:
+  title: IO Sign - Backend API
+  version: 0.0.1
+servers:
+  - url: https://app-backend.io.italia.it/api/v1/fci
```
gunzip: why fci and not 'sign'?

docs/openapi-backend.yaml
```diff
@@ -0,0 +1,393 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/QtspClausesMetadata"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
+
+  /qtsp/clauses/filled_document:
```
gunzip, line 56:
```suggestion
  /qtsp/clauses/filled-document:
```

docs/openapi-backend.yaml
grausof: Because app side "firma con Io" has always been called fci.

@hevelius

hevelius: On the app side from the beginning `fci` was used to indicate the feat flag, file name convention and other references. It's not a problem to change this url to `sign` if you want maybe can make sense because it refers to io-sign be.

lucacavallaro: yes, @hevelius please change it to `sign` (or `io-sign` when `sign` is ambiguous)

the BE endpoint should be `/api/v1/sign/`

docs/openapi-backend.yaml
lucacavallaro: @gunzip @grausof  if `filled_document` is a property it should be written in `snake_case`, if it is a collection it is ok to leave it in the `kebab-case`

lucacavallaro (approved):

gunzip (approved):

gunzip (approved):

PR #29, opened 2022-10-24
docs: add new openapi for issuer

lucacavallaro:
#### List of Changes

Proposal of stable APIs for Issuers

- `snake_case` 🎉
- `Product` is now `Dossier`
-  `Clause` now have a `type` field that can be `unfair`, `required` or `optional`
-  New timestamps on `Signature Request`
- New endpoint `POST /signers` to get `Signer` objects by `fiscal code`
- The issuer now can specify `coordinates`, `page` and `size` in order to create a new `signature field` into the PDF document

#### Types of changes

- [x] Chore (nothing changes by a user perspective)

#### Checklist:


docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,570 @@
...
+      type: object
+      properties:
+        title:
+          type: string
+        signature_fields:
+          type: array
+          items:
+            oneOf:
+              - $ref: "#/components/schemas/SignatureField"
+              - $ref: "#/components/schemas/SignatureFieldToBeCreated"
```
grausof:
```suggestion
          items:
            $ref: "#/components/schemas/SignatureField"
```

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,570 @@
...
+        page:
+          type: number
+          minimum: 0
+        size:
+          type: object
+          properties:
+            w: { type: number, minimum: 0 }
+            h: { type: number, minimum: 0 }
+        clause:
+          $ref: "#/components/schemas/Clause"
```
grausof, line 381:
```suggestion
```

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,570 @@
...
+    SignatureField:
+      type: object
+      properties:
+        unique_name:
+          type: string
+        clause:
+          $ref: "#/components/schemas/Clause"
+      required:
+        - unique_name
+        - clause
```
grausof:
```suggestion
      properties:
        field_detail:
          oneOf: 
            - $ref: "#/components/schemas/FieldToBeCreated"
            - $ref: "#/components/schemas/FieldUniqueName"
        clause:
          $ref: "#/components/schemas/Clause"
      required:
        - field_detail
        - clause
```

grausof: I prefer to have all the information relating to a field (coordinates, dimensions and page) in a single object such as signature_field_detail

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,570 @@
...
+        id:
+          $ref: "#/components/schemas/Id"
+        dossier_id:
+          $ref: "#/components/schemas/Id"
+        expires_at:
+          $ref: "#/components/schemas/Timestamp"
+        documents:
+          type: array
+          items:
+            oneOf:
```
grausof, line 582:
```suggestion
            anyOf:
```

grausof (changes requested):

docs/openapi-issuer.yaml
lucacavallaro: ok! I'll try to group them in order to make these properties easier to use

docs/openapi-issuer.yaml
lucacavallaro: why `anyOf`?

grausof: `documents` can match one or more subschemas. There can be DocumentToBeUploaded and also DocumentUploaded in the same array

lucacavallaro: yes, but that is the purpose of `oneOf` (if I understood correctly)

for each element it checks if the shape matches with one of the scheme provided

`anyOf` checks the shape of an object between one and more schemes

hevelius (commented):
LGTM / I don't know perfectly the specs for issuer but I follow the comments of @grausof

docs/openapi-issuer.yaml
grausof: Mhh... yes you are right!

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,570 @@
...
+            schema:
+              $ref: "#/components/schemas/CreateDossierBody"
+      responses:
+        "201":
+          description: The Dossier detail
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/DossierDetailView"
+        "400":
```
gunzip, line 30: All API that pass though the API management must handle: 400, 401, 403, 429.

see https://github.com/pagopa/io-functions-services/blob/master/openapi/index.yaml#L296

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,570 @@
...
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
+        default:
+          $ref: "#/components/responses/Unexpected"
+
+  /signers:
+    post:
+      operationId: getSignerByFiscalCode
```
gunzip, line 75: why is this a post and not a get?

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,570 @@
...
+            schema:
+              $ref: "#/components/schemas/GetSignerByFiscalCodeBody"
+      responses:
+        "201":
+          description: The Signer detail
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SignerDetailView"
+        "400":
```
gunzip, line 92: what about 404?

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,570 @@
...
+      summary: Set the status of a Signature Request
+      parameters:
+        - in: path
+          name: id
+          required: true
+          schema:
+            $ref: "#/components/schemas/Id"
+      responses:
+        "204":
+          description: Signature Request status successfully setted
```
gunzip:
```suggestion
          description: Signature Request status successfully set
```

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,570 @@
...
+        - detail
+
+    Clause:
+      type: object
+      properties:
+        title:
+          type: string
+        type:
+          type: string
+          enum: ["REQUIRED", "UNFAIR", "OPTIONAL"]
```
gunzip, line 344:
```suggestion
          x-extensible-enum: ["REQUIRED", "UNFAIR", "OPTIONAL"]
```

gunzip (dismissed):

docs/openapi-issuer.yaml
gunzip: anyOf is not supported by the codegen

docs/openapi-issuer.yaml
gunzip: plase add documentation for these fields

lucacavallaro: why `x-extensibile-enum` ?

docs/openapi-issuer.yaml
lucacavallaro: in order to avoid passing the fiscal code as route param

grausof: What's the problem with passing the fiscalCode as route param?

lucacavallaro: because the fiscal code is sensitive data (pii) and I would like to avoid "involuntary logging"

docs/openapi-issuer.yaml
gunzip: we can use enum if we're absolutely certain that this list won't change in future.

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,627 @@
...
+      operationId: putSignatureRequestStatus
+      tags:
+        - Signature Request
+      summary: Set the status of a Signature Request
+      parameters:
+        - in: path
+          name: id
+          required: true
+          schema:
+            $ref: "#/components/schemas/Id"
```
grausof, line 174: there is no requestBody. How can I specify the status?

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,627 @@
...
+        - in: path
+          name: req_id
+          required: true
+          schema:
+            $ref: "#/components/schemas/Id"
+        - in: path
+          name: doc_id
+          required: true
+          schema:
+            $ref: "#/components/schemas/Id"
```
grausof, line 215: Isn't it better to specify?
```
req_id -> signature_request_id
doc_id -> document_id
```

docs/openapi-issuer.yaml
lucacavallaro: good catch! :D

docs/openapi-issuer.yaml
lucacavallaro: Given the rest of the used path, I think their "meaning" is clear

grausof:
⚠️ `gen-api-models` crashes using these specifications
```
UploadUrl -> src/infra/api-models/UploadUrl.ts
Error: SyntaxError: Expression expected. (59:7)
  57 |       
  58 |       
> 59 |       });
     |       ^
  60 | 
  61 |       export const SignatureField = t.exact(
  62 |         t.intersection(
error Command failed with exit code 1.
```

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,635 @@
...
+            w: { type: number, minimum: 0 }
+            h: { type: number, minimum: 0 }
+
+    SignatureField:
+      type: object
+      properties:
+        attrs:
+          oneOf:
+            - $ref: "#/components/schemas/SignatureFieldAttributes"
+            - $ref: "#/components/schemas/SignatureFieldToBeCreatedAttributes"
```
grausof:
```suggestion
        attrs:
          type: object
          oneOf:
            - $ref: "#/components/schemas/SignatureFieldAttributes"
            - $ref: "#/components/schemas/SignatureFieldToBeCreatedAttributes"
```

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,635 @@
...
+          $ref: "#/components/schemas/Id"
+        signer_id:
+          allOf:
+            - $ref: "#/components/schemas/Id"
+            - example: "01BX5ZZKBKACTAV9WEVGEMMVS0"
+        expires_at:
+          $ref: "#/components/schemas/Timestamp"
+      required:
+        - dossier_id
+        - signer_id
```
grausof: Here we lack the forwarding mode whether by message or qrcode.

lucacavallaro: yep

grausof: we leave `send_message: true` ?

lucacavallaro: `notify_user_via_io_message`

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,639 @@
...
+
+    SignatureRequest:
+      type: object
+      properties:
+        id:
+          $ref: "#/components/schemas/Id"
+        dossier_id:
+          $ref: "#/components/schemas/Id"
+        expires_at:
+          $ref: "#/components/schemas/Timestamp"
```
grausof: `signer_id` is missing

lucacavallaro: After giving it some thought, I decided it wasn't necessary to include `signer_id` in this response, maybe `dossier_id` is redundant either. Do you believe it should be included?

grausof: I think it is useful as the `ISSUER` given a `signature_request_id` will no longer be able to trace the `signer_id` if he has not stored this information somewhere.

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,639 @@
...
+
+    CreateSignatureRequestBody:
+      type: object
+      properties:
+        dossier_id:
+          $ref: "#/components/schemas/Id"
+        signer_id:
+          allOf:
+            - $ref: "#/components/schemas/Id"
+            - example: "01BX5ZZKBKACTAV9WEVGEMMVS0"
```
grausof:
```suggestion
        signer_id:
          $ref: "#/components/schemas/Id"
```

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,639 @@
...
+        id:
+          $ref: "#/components/schemas/Id"
+        dossier_id:
+          $ref: "#/components/schemas/Id"
+        expires_at:
+          $ref: "#/components/schemas/Timestamp"
+        documents:
+          type: array
+          items:
+            oneOf:
```
grausof, line 582:
```suggestion
          items:
            type: array
            oneOf:
```

docs/openapi-issuer.yaml
grausof: typo?

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,639 @@
...
+            h: { type: number, minimum: 0 }
+
+    SignatureField:
+      type: object
+      properties:
+        attrs:
+          type: object
+          oneOf:
+            - $ref: "#/components/schemas/SignatureFieldAttributes"
+            - $ref: "#/components/schemas/SignatureFieldToBeCreatedAttributes"
```
grausof, line 379:
```suggestion
    SignatureField:
      properties:
        attrs:
          oneOf:
            - $ref: "#/components/schemas/SignatureFieldAttributes"
            - $ref: "#/components/schemas/SignatureFieldToBeCreatedAttributes"
```
In this way the specifications are validated!

docs/openapi-issuer.yaml
lucacavallaro: nope: it a trick to provide an example with different ID

grausof: I don't think this trick is necessary. The id is always generated on the server side so it cannot confuse the issuer. Furthermore, this type of structure does not make the generator work. 
CC @gunzip

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,639 @@
+openapi: 3.0.3
+info:
+  title: Firma con IO - Issuer API
+  version: 1.0.0
+servers:
+  - url: https://api.io.pagopa.it/api/v1/sign
+    description: production
+security:
+  - SubscriptionKey: []
+paths:
```
gunzip, line 10: not urgent but you probably want to create "Links" sections with the returned URLs to store (upload) documents: https://swagger.io/docs/specification/links/

grausof (approved):
LGTM

docs/openapi-issuer.yaml
```diff
@@ -0,0 +1,639 @@
...
+      properties:
+        attrs:
+          type: object
+          oneOf:
+            - $ref: "#/components/schemas/SignatureFieldAttributes"
+            - $ref: "#/components/schemas/SignatureFieldToBeCreatedAttributes"
+        clause:
+          $ref: "#/components/schemas/Clause"
+      required:
+        - unique_name
```
grausof, line 383: `attrs` and not `unique_name`

lucacavallaro: check out the latest spec https://github.com/pagopa/io-sign/blob/functions-refactor/apps/io-func-sign-issuer/openapi.yaml

PR #30, opened 2022-10-31
docs: Add notification endpoint [SFEQS-1113]

grausof:
This PR adds a new entity (notification) to send the signature request via IO message

#### Types of changes
- [X] New feature (non-breaking change which adds functionality)

#### Checklist:
- [X] My change requires a change to the documentation.

### DIFF
--- a/docs/openapi-issuer.yaml
+++ b/docs/openapi-issuer.yaml
@@ -233,7 +233,7 @@ paths:
         default:
           $ref: "#/components/responses/Unexpected"
 
-  /notification:
+  /notifications:
     post:
       operationId: sendNotification
       tags:
@@ -645,12 +645,21 @@ components:
             qr_code_url:
               type: string
               format: uri
-            notified_via_io_message:
-              type: boolean
-              default: false
+            notification:
+              type: object
+              properties:
+                notified_via_io_message:
+                  type: boolean
+                  default: false
+                io_message_id:
+                  type: string
+              required:
+                - notified_via_io_message
+
           required:
             - status
             - qr_code_url
+            - notification
 
     SignatureRequestDetailView:
       oneOf:
@@ -679,7 +688,7 @@ components:
     NotificationDetailView:
       type: object
       properties:
-        id:
-          $ref: "#/components/schemas/Id"
+        io_message_id:
+          type: string
       required:
-        - id
+        - io_message_id
--- a/docs/openapi-issuer.yaml
+++ b/docs/openapi-issuer.yaml
@@ -69,6 +69,35 @@ paths:
           $ref: "#/components/responses/TooManyRequests"
         default:
           $ref: "#/components/responses/Unexpected"
+  /notification:
+    post:
+      operationId: sendNotification
+      tags:
+        - Notification
+      summary: Send a signature request notification to user via IO message
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/SendNotificationBody"
+      responses:
+        "201":
+          description: The notification detail
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/NotificationDetailView"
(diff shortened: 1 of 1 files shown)


docs/openapi-issuer.yaml
```diff
@@ -233,6 +233,36 @@ paths:
         default:
           $ref: "#/components/responses/Unexpected"
 
+  /notification:
```
gunzip: Isn't the notification sent automatically?

grausof: No, we have decided to make this operation manual in such a way that the issuer can receive an error message if the receipt of messages by the user is not enabled. We have switched to using only the standard service without opt-in.

gunzip:
```suggestion
  /notifications:
```

gunzip: What about:

`/signature-requests/{id}/notifications` ?

docs/openapi-issuer.yaml
```diff
@@ -618,6 +645,9 @@ components:
             qr_code_url:
               type: string
               format: uri
+            notified_via_io_message:
```
gunzip: What about storing message id here instead, ie.

`...signatureRequest, 
  notifications: { 
     io_message_id: <id>
  }`

This would let (one day) the issuer recover the IO message notification status.

docs/openapi-issuer.yaml
```diff
@@ -233,6 +233,36 @@ paths:
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/SendNotificationBody"
+      responses:
+        "201":
+          description: The notification detail
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/NotificationDetailView"
```
gunzip, line 254: What's the nofification id?

docs/openapi-issuer.yaml
grausof: It seems right!

docs/openapi-issuer.yaml
grausof: is the `io_message_id`

docs/openapi-issuer.yaml
grausof: I think that `notifications` is not a property of the signature request. I'd rather have separated it

docs/openapi-issuer.yaml
grausof: i renamed `id` to `io_message_id`

docs/openapi-issuer.yaml
gunzip: @lucacavallaro what about this?

lucacavallaro: 1. the issuer can only send **1** `IO Message` per `Signature Request` .
2. the `IO Message IO` must be stored

I propose to add a property called `notification` to `Signature Request` (yes, "_notification_" not "notifications" since the notification is only one) and a related `PUT /signature-requests{id}/notification` to update it (when it's empty)

docs/openapi-issuer.yaml
```diff
@@ -233,6 +233,36 @@ paths:
...
+          schema:
+            $ref: "#/components/schemas/Id"
+      responses:
+        "201":
+          description: The notification detail
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/NotificationDetailView"
+        "400":
```
gunzip, line 258: what if you call this endpoint twice?

grausof: You get an error because the message has already been sent!

gunzip: with which status code? I'd use `409 confict` in case.

grausof: It seems right.  I add `409`.

lucacavallaro (approved):

PR #31, opened 2022-10-31
core: Add PDF content validator [SFEQS-1110]

grausof:
This PR adds the validation of the pdf files uploaded through the [pdf-lib](https://pdf-lib.js.org/) library.

#### Motivation and Context
Currently the file upload flow does not provide any control on the type and / or format therefore it is possible to upload any type of file without any control. File format control is added with this PR.

Sested on simulator

#### Types of changes
- [X] Chore (nothing changes by a user perspective)

#### Checklist:
- [X] I have updated the documentation accordingly.


src/infra/azure/storage/document.ts
```diff
@@ -25,3 +27,14 @@ export const makeDeleteDocumentUploaded: R.Reader<
...
   pipe(client, selectBlob(documentId), blobDelete);
+
+export const makeDownloadDocumentUploaded: R.Reader<
+  ContainerClient,
+  DownloadUploadDocumentFromBlob
+> = (client) => (documentId) =>
+  pipe(client, selectBlob(documentId), (blob) =>
+    TE.tryCatch(
+      () => blob.downloadToBuffer(),
+      () => new Error("The blob cannot be downloaded.")
```
gunzip: remember to log the original error at least

src/app/use-cases/validate-pdf.ts
```diff
@@ -0,0 +1,16 @@
...
+import { pipe } from "fp-ts/function";
+
+export const getPdfMetadata = (buffer: Buffer) =>
+  pipe(
+    TE.tryCatch(
+      () =>
+        PDFDocument.load(buffer, {
+          updateMetadata: false,
+        }),
+      () => new Error("This is not a valid PDF")
```
gunzip: can we keep some explicit error here? ie.

gunzip:
```suggestion
      toError
```

src/app/use-cases/status-document.ts
```diff
@@ -53,6 +53,8 @@ const whenValidationInProgress =
       case "MARK_AS_READY":
         return E.right({ ...document, status: "READY" });
       case "MARK_AS_INVALID":
+        // eslint-disable-next-line functional/immutable-data
```
gunzip, line 56:
```suggestion
        // eslint-disable-next-line @typescript-eslint/no-unused-vars
        const { url, ...doc } = document;
        return E.right({ ...doc, status: "VALIDATION_ERROR" });
```

grausof: I get the following error: `'url' is assigned a value but never used.`

src/app/use-cases/validate-pdf.ts
grausof: 👍

src/app/use-cases/status-document.ts
gunzip: I've updated the suggestion

